PR #51, opened 2020-02-04
Jb/experimental config

jackbackrack:
Put in experimental config mechanism for new compiler features.

### DIFF
--- a/compiler/stz-config.stanza
+++ b/compiler/stz-config.stanza
@@ -36,9 +36,10 @@ defsyntax stanza-config :
   defrule long! = (?s:#long) : s
   fail-if long! = () : PE(closest-info(), "Expected a long here.")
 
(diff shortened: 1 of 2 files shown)


compiler/stz-config.stanza
```diff
@@ -36,6 +36,11 @@ defsyntax stanza-config :
   defrule long! = (?s:#long) : s
   fail-if long! = () : PE(closest-info(), "Expected a long here.")
 
+  defproduction bool! : True|False
+  defrule bool! = (?v:true) : true
```
CuppoJava: defrule bool! = (?v) when unwrap-token(v) is True|False : unwrap-token(v)

CuppoJava (commented):
Please fix bool parser.

PR #53, opened 2020-02-04
optimize dispatch code gen to remove extra branches based on types of…

jackbackrack:
Speed up dispatch by removing unnecessary branches in tree gen.

### DIFF
--- a/compiler/stz-stitcher.stanza
+++ b/compiler/stz-stitcher.stanza
@@ -496,7 +496,9 @@ public defn Stitcher (packages:Collection<VMPackage>, bindings:Bindings|False, s
           ;Test bits
           val marker-branches = unique-id(stubs)
           val ref-branches = unique-id(stubs)
-          E $ AndL(BITS, object, INT(7))      
+          val [p?, m?, r?] = [not empty?(prim-targets), not empty?(marker-targets), not empty?(ref-targets)]
+          if p? or r? :
+            E $ AndL(BITS, object, INT(7))
           for entry in prim-targets do :
             val x = key(entry)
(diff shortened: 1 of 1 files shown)


CuppoJava:
```
          ;Checking for primitives or references requires knowing the tag bits.
          ;So create the tag bits if we have either case.
          if p? or r? :
            E $ AndL(BITS, object, INT(7))

          ;Jump to the appropriate primitive target if the object
          ;is one of the given primitives.
          for entry in prim-targets do :
            val x = key(entry)
            val target = value(entry)
            val tagbits =
              if x == id-indices[CORE-BYTE-ID] : INT(3)
              else if x == id-indices[CORE-CHAR-ID] : INT(4)
              else if x == id-indices[CORE-INT-ID] : INT(0)
              else if x == id-indices[CORE-FLOAT-ID] : INT(5)
              else : fatal("Not a primitive type")
            E $ BreakL(target, EqOp(), BITS, tagbits)

          ;At this point, we are guaranteed not to branch to one of the primitive targets.
          ;We now need to decide whether to branch to a marker or a ref target.
          ;Perform this check only if we actually have both marker and ref targets.
          if m? and r? :
            E $ BreakL(M(marker-branches), EqOp(), BITS, INT(2))

          ;At this point, we are guaranteed not to branch to a primitive or marker target.
          ;If we need to branch to a ref target, then we need to check its tagbit in order
          ;to safely access its heap tag.
          if r? :
            E $ BreakL(M(ref-branches), EqOp(), BITS, INT(1))

          ;?????
          if (p? and not m?) or r? :
            E $ Goto(default-target)
```

PR #54, opened 2020-02-06
Jb/el do

jackbackrack:
Create an optimized do implementation that doesn't call map.

### DIFF
--- a/compiler/stz-el-ir.stanza
+++ b/compiler/stz-el-ir.stanza
@@ -1154,11 +1154,6 @@ public deftype ELItem :
   EClosure <: ELItem
   ELBigItem <: ELItem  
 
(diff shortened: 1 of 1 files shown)


jackbackrack:
i'm wondering if there's a more robust way to do this.  for example, perhaps this can be done by copying the map but adding a false and then using dead code elimination to prune out all the unnecssary object creation.  this hasn't shown much in the way of speed ups so far so it's inclusion should perhaps be delayed.

CuppoJava:
It can be done automatically by specializing functions based upon the context in which their return value is used. In this case, the compile could derive a version of `map` that is used whenever it can deduce that its return value is not used. Because the return value is not used, then it does not need to create any output objects. Once that is recursively applied to everything in `map`, it will be identical to `do`.

PR #55, opened 2020-02-06
Jb/stack checks

jackbackrack:
remove stack check if function is small and doesn't have any non-tail-calls.

### DIFF
--- a/compiler/stz-compiler-main.stanza
+++ b/compiler/stz-compiler-main.stanza
@@ -295,10 +295,6 @@ public defn compile (proj-manager:ProjManager,
     for f in funcs(vmpackage(npkg)) do :
       emit(emitter, LinkLabel(id(f)))
       allocate-registers(func(f), emitter, backend, stubs, false)
-    val sizes = get-smap-sizes()
-    val keys = qsort(keys(sizes))
-    for key in keys do :
-      println("/* %_ => %_ */" % [key, sizes[key]])
 
   defn emit-all-system-stubs (stitcher:Stitcher, stubs:AsmStubs) :
     val emitter = file-emitter()
--- a/compiler/stz-compiler-main.stanza
+++ b/compiler/stz-compiler-main.stanza
@@ -179,8 +179,17 @@ public defn compile (proj-manager:ProjManager,
     switch(situation) :
       `optimized-asm :
         val packages = Vector<VMPackage|StdPkg>()
+        val opt-timer = MillisecondTimer("Opt")
+        start(opt-timer)
         add(packages, combine-and-lower(/packages(result) as Tuple<EPackage|FastPkg>))
+        stop(opt-timer)
+        println(opt-timer)
+        val backend-timer = MillisecondTimer("Backend")
+        start(backend-timer)
         compile-vmpackages({false}, to-tuple(packages), bindings(result), output as String, false)
+        stop(backend-timer)
+        println(backend-timer)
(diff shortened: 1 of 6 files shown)


compiler/stz-reg-alloc.stanza
```diff
@@ -204,6 +204,8 @@ defpackage stz/reg-alloc :
 ;===================== Driver ===============================
 ;============================================================
 
+val SMALL-SMAP-SIZE = 16 ;; SIZE OF SMAP WHICH CAN BE CALLED WITHOUT CHECKING STACK-SIZE
```
CuppoJava: Create a 
```
;====================================
;========== PARAMETERS ===============
;====================================
```

section at the top of the file and put this definition there. 

Please follow code comment conventions. Comments start with a single semi-colon are capitalized and then lower-case.

compiler/stz-reg-alloc.stanza
```diff
@@ -2681,6 +2683,31 @@ defn assemble (emitter:CodeEmitter, stackmap:StackMap, stubs:AsmStubs) :
         val amb-branch = label-table[amb(e)]
         E $ asm-MethodDispatch(multi(e), num-header-args(e), no-branch, amb-branch)        
       (e) : fatal("Not yet implemented: %_" % [e])
+
```
CuppoJava, line 2705: This logic is too fragile. Let's introduce a new instruction called:

```
SectionMarker :
  name: Symbol
```

And have the preamble be surrounded by a pair of `SectionMarker` instructions. Then strip away everything in that section.

compiler/stz-reg-alloc.stanza
```diff
@@ -2436,7 +2438,7 @@ defn assemble (emitter:CodeEmitter, stackmap:StackMap, stubs:AsmStubs) :
           (t:ExtendStack) :
             val return-lbl = unique-id(stubs)
             E $ SetL(R0, LM(return-lbl))
-            E $ SetL(R1, INT(size(stackmap) + 8))
+            E $ SetL(R1, INT(size(stackmap) + (8 + SMALL-SMAP-SIZE)))
```
CuppoJava, line 2460: This might be the right logic, but before I can merge it I need to refresh my head about the details of the stack layout. It's tricky, and if we get it wrong, it manifests as random crashes every few weeks, so it's important we get it right on the first try.

To do that, I need to know what is the stack layout, what is enforced by this instruction, why that `+8` is there, etc.

compiler/stz-compiler-main.stanza
```diff
@@ -179,6 +179,7 @@ public defn compile (proj-manager:ProjManager,
     switch(situation) :
       `optimized-asm :
         val packages = Vector<VMPackage|StdPkg>()
+        val opt-timer = MillisecondTimer("Opt")
```
CuppoJava: Please cleanup before merge.

CuppoJava (changes requested):
This is a useful but tricky request to merge. Will need time to review.

compiler/stz-reg-alloc.stanza
jackbackrack: done.

compiler/stz-reg-alloc.stanza
jackbackrack: the +8 is yours.  i just added SMALL-SMAP-SIZE.

compiler/stz-reg-alloc.stanza
jackbackrack: done.

PR #58, opened 2020-02-06
fix bug in dispatch to compare etc

jackbackrack:
fix bug that prevented int compare method from being resolved.

### DIFF
--- a/compiler/stz-dispatch-dag.stanza
+++ b/compiler/stz-dispatch-dag.stanza
@@ -484,11 +484,12 @@ public defn all-solns (dag:Dag, case:Tuple<Arg>, include-amb?:True|False) -> Tup
   ;Algorithm
   defn* loop (e:DagEntry) :
     val arg = case[depth(e)]
-    val hits = for e in entries(e) filter : overlap?(key(e), arg)
-    if empty?(hits) :
-      loop(default(e))
-    else :
-      for hit in hits do : loop(value(hit))
+    if not bottom?(arg) :
+      val hits = for e in entries(e) filter : overlap?(key(e), arg)
+      if empty?(hits) :
+        loop(default(e))
+      else :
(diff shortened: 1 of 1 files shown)


CuppoJava:
Thanks for the getting to the bottom of this. 

Can you give an example and explanation of the problem and the fix for it?

jackbackrack:
The example is:

public defmulti less?<?T> (a:Comparable<?T>, b:T) -> True|False
defmethod less?<?T> (a:Comparable<?T>, b:T) : compare(a, b) < 0
...
defmethod less? (a:Int, b:Int) -> True|False :  ($prim lt a b)

defn f (a:Int, b:Int) : a < b

the call to less? in f will result in two applicable methods.  the current dispatch code:

  defn* loop (e:DagEntry) :
    val arg = case[depth(e)]
    if not bottom?(arg) :
      for e in entries(e) do :
        loop(value(e)) when overlap?(key(e), arg)
      loop(default(e))

the dagentry will contain int method as an entry and the comparable method as a default.
when running the current dispatch code it will return both methods.  in the modified code i only used the default if there are no other matching methods:

  defn* loop (e:DagEntry) :
    val arg = case[depth(e)]
    val hits = for e in entries(e) filter : overlap?(key(e), arg)
    if empty?(hits) :
      loop(default(e))
    else :
      for hit in hits do : loop(value(hit))

this then successfully dispatches to the int method in the example.

i think i should also put in the bottom guard:

  defn* loop (e:DagEntry) :
    val arg = case[depth(e)]
    if not bottom?(arg) :
      val hits = for e in entries(e) filter : overlap?(key(e), arg)
      if empty?(hits) :
        loop(default(e))
      else :
        for hit in hits do : loop(value(hit))

CuppoJava:
Thanks for the pull request. Changed detection logic to use `subarg?` predicate and merged into `0.13.47`.

PR #59, opened 2020-02-06
Jb/faster dos

jackbackrack:
add tuple do method and
use only a single index variable instead of two.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -5253,11 +5253,11 @@ defmethod get (a:CharArray, r:Range) -> String :
 ;                     =======
 
 defmethod do (f: Char -> ?, xs:CharArray) :
-   defn* loop (i:Int, n:Int) :
-      if n > 0 :
+   val n = length(xs)
+   let loop (i:Int = 0) :
+      if i < n :
          f(xs[i])
-         loop(i + 1, n - 1)
-   loop(0, length(xs))
+         loop(i + 1)
 
 ;============================================================
 ;==================== Block Copying =========================
@@ -5930,18 +5930,17 @@ public defn reverse (r:Range) :
 
 defmethod do (f: Int -> ?, r:Range&Lengthable) :
    val s = step(r)
-   defn* loop (i:Int, n:Int) :
-      if n > 0 :
+   val n = length(r)
+   let loop (i:Int = 0) :
+      if i < n :
          f(i)
-         loop(i + s, n - 1)
-   loop(start(r), length(r))
+         loop(i + s)
 
 defmethod do (f: Int -> ?, r:Range) :
    val s = step(r)
-   defn* loop (i:Int) :
+   let loop (i:Int = start(r)) :
       f(i)
       loop(i + s)
-   loop(start(r))
 
 ;                  Inclusion
 ;                  =========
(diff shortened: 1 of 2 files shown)


core/core.stanza
```diff
@@ -6296,34 +6294,40 @@ defmethod do<?T,?S,?U> (f:(T,S,U) -> ?, xs:Seqable<?T>, ys:Seqable<?S>, zs:Seqab
...
+   let loop (i:Int = 0) :
+      if i < n :
          f(xs[i])
-         loop(i + 1, n - 1)
-   loop(0, length(xs))
+         loop(i + 1)
 
 defmethod do<?T> (f:T -> ?, xs:List<?T>) -> False :
-   defn* loop (xs:List<T>) :
+   let loop (xs:List<T> = xs) :
```
CuppoJava, line 6327: Thanks!

core/core.stanza
```diff
@@ -6296,34 +6294,40 @@ defmethod do<?T,?S,?U> (f:(T,S,U) -> ?, xs:Seqable<?T>, ys:Seqable<?S>, zs:Seqab
...
+      val n = length(xs)
+      let loop (i:Int = 0) :
+         if i < n :
+            f(xs[i])
+            loop(i + 1)
 
 defmethod do (f: Char -> ?, xs:String) :
-   defn* loop (i:Int, n:Int) :
-      if n > 0 :
+   val n = length(xs)
```
CuppoJava, line 6320: Verify that closing over `n` doesn't make it slower.

core/core.stanza
```diff
@@ -6296,34 +6294,40 @@ defmethod do<?T,?S,?U> (f:(T,S,U) -> ?, xs:Seqable<?T>, ys:Seqable<?S>, zs:Seqab
...
 defmethod do<?T> (f: T -> ?, xs:Array<?T>) :
-   defn* loop (i:Int, n:Int) :
-      if n > 0 :
+   val n = length(xs)
+   let loop (i:Int = 0) :
+      if i < n :
+         f(xs[i])
+         loop(i + 1)
+
+defmethod do<?T> (f: T -> ?, xs:Tuple<?T>) :
```
CuppoJava, line 6303: Thanks!

core/core.stanza
```diff
@@ -5930,18 +5930,16 @@ public defn reverse (r:Range) :
...
+   let loop (i:Int = start(r), n:Int = length(r)) :
       if n > 0 :
          f(i)
          loop(i + s, n - 1)
-   loop(start(r), length(r))
 
 defmethod do (f: Int -> ?, r:Range) :
    val s = step(r)
-   defn* loop (i:Int) :
+   let loop (i:Int = start(r)) :
```
CuppoJava, line 5940: Thanks!

core/core.stanza
```diff
@@ -5930,18 +5930,16 @@ public defn reverse (r:Range) :
 
 defmethod do (f: Int -> ?, r:Range&Lengthable) :
    val s = step(r)
-   defn* loop (i:Int, n:Int) :
+   let loop (i:Int = start(r), n:Int = length(r)) :
```
CuppoJava, line 5933: Thanks!

core/core.stanza
```diff
@@ -5253,11 +5253,11 @@ defmethod get (a:CharArray, r:Range) -> String :
 ;                     =======
 
 defmethod do (f: Char -> ?, xs:CharArray) :
-   defn* loop (i:Int, n:Int) :
-      if n > 0 :
+   val n = length(xs)
```
CuppoJava, line 5256: Verify that closing over `n` doesn't make it slower. It might be better to do this:

```
let loop (i:Int = 0, n:Int = length(xs)) :
  if i < n :
    f(xs[i])
    loop(i + 1, n)
```

CuppoJava (changes requested):
Thanks! Can you do a quick timing check to make sure that closing over a variable doesn't make the loop slower? (Or look at the generated code.)

core/core.stanza
jackbackrack: no n is not closed over as the compiler inlines the function!

jackbackrack:
i have extensive timings to report that this change alone speeds up the compiler by 15%.  i also verified that the loop is inlined and that n is not closed over anymore.  the compiler compile time goes from 98s to 85s.

jackbackrack:
Also i did a full bootstrap of the compiler compiling itself four times in succession.

CuppoJava:
Awesome. Thanks!

PR #61, opened 2020-02-21
Jb/sorted sets

jackbackrack:
Use sorted ints as representation of Nums so that all set operations will be O(n) instead of O(n^2).  speeds up compiler by 18%.


CuppoJava:
(Mostly) Merged into Stanza `0.13.51`. Removed the NumSet abstraction entirely, and instead just maintained sorted invariant of the stz/dispatch-dag/Nums datastructure. 

The conservative `default-reachable?` implementation was not merged, and was instead replaced with a `subarg-of-union?` predicate. Can you verify whether this is as fast as it needs to be?

CuppoJava:
Merged into `0.13.52`. Verified that it avoids slowdown for `default-reachable?` predicate.

PR #63, opened 2020-02-22
type raw array and add precise inference rule for objectget and newob…

jackbackrack:
Typed raw array so that in optimized mode the element types are propagated properly.  Added a couple type inference rules so that the element types and object types are precisely inferred.

### DIFF
--- a/compiler/stz-el.stanza
+++ b/compiler/stz-el.stanza
@@ -1866,6 +1866,7 @@ defn annotate-body-instructions (e:EBody, state0:Table<Int,EType>, vt:VarTable) 
       (i:EVoidTuple) : trusted(x(i))
       (i:ETupleGet) : untrusted(x(i))
       (i:EObject) : trusted(x(i))
+      (i:EObjectGet) : trusted(x(i))
       (i:EArray) : trusted(x(i))
       (i:EStruct) : trusted(x(i))
       (i:EPtr) : trusted(x(i))
@@ -1873,7 +1874,6 @@ defn annotate-body-instructions (e:EBody, state0:Table<Int,EType>, vt:VarTable) 
       (i:EInterpret) : untrusted(x(i))
       (i:EConv) : untrusted(x(i))
       (i:EPrim) : trusted(x(i))
-      (i:ENewObject) : trusted(x(i))
       (i:ECall) :
         match(x(i)) :
           (x:EVarLoc) :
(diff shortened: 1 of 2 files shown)


CuppoJava:
Merged, the RawArray parametric types, but still need to review the changes to the type inference.

compiler/stz-el.stanza
```diff
@@ -1866,14 +1866,14 @@ defn annotate-body-instructions (e:EBody, state0:Table<Int,EType>, vt:VarTable)
...
       (i:EObject) : trusted(x(i))
+      (i:EObjectGet) : trusted(x(i))
       (i:EArray) : trusted(x(i))
       (i:EStruct) : trusted(x(i))
       (i:EPtr) : trusted(x(i))
       (i:ELoad) : untrusted(x(i))
       (i:EInterpret) : untrusted(x(i))
       (i:EConv) : untrusted(x(i))
       (i:EPrim) : trusted(x(i))
-      (i:ENewObject) : trusted(x(i))
```
CuppoJava: What was the reasoning for moving this out of this section and into the next?

compiler/stz-el.stanza
```diff
@@ -1866,14 +1866,14 @@ defn annotate-body-instructions (e:EBody, state0:Table<Int,EType>, vt:VarTable)
       (i:EVoidTuple) : trusted(x(i))
       (i:ETupleGet) : untrusted(x(i))
       (i:EObject) : trusted(x(i))
+      (i:EObjectGet) : trusted(x(i))
```
CuppoJava, line 1869: Looks good.

compiler/stz-el.stanza
jackbackrack: i tested and found in fact it doesn't affect the results to leave it the way it was.  the important thing is to have rawarray be typed it appears.

compiler/stz-el.stanza
```diff
@@ -89,6 +89,8 @@ defn lower (epackage:EPackage, optimize?:True|False) -> EPackage :
   run-pass("Lift Closures", lift-closures, "closurelifted", false)
   run-pass("Lift Type Objects", lift-type-objects, "typelifted", false)
 
+  spit("out.txt", cur-package)
```
CuppoJava: Can you get rid of the debug output?

CuppoJava (changes requested):

jackbackrack:
done.

PR #68, opened 2020-02-26
speed up dispatch-dag default to use conservative check for speed

jackbackrack:
revert back to conservative dispatch-dag default method to improve speed of compiler.


CuppoJava:
Merged into `0.13.52`

PR #71, opened 2020-05-06
use unique filenames and cleanup processes so multiple processes can be used

jackbackrack:


core/core.stanza
```diff
@@ -7260,8 +7261,11 @@ public lostanza defn sleep-us (us:ref<Long>) -> ref<False> :
 ;                   Process Structure
 ;                   =================
 
+lostanza var next-pipe-id:int = 0
```
CuppoJava: Use capital letters for global variables.

runtime/driver.c
```diff
@@ -835,15 +836,22 @@ void initialize_launcher_process (){
   }
 }
 
+int delete_process_pipes (FILE* input, FILE* output, FILE* err) {
```
CuppoJava: Check that the input/output/errs are actually created before closing. Inspect their creation in launch_process.

runtime/driver.c
```diff
@@ -835,15 +836,22 @@ void initialize_launcher_process (){
...
+  fclose(input);
+  fclose(output);
+  fclose(err);
```
CuppoJava: Check for error return code. Return the return code upon error.

core/core.stanza
```diff
@@ -7425,6 +7430,10 @@ defn SystemCallException (msg:String) :
     val TERMINATED = 2
     val STOPPED = 3
 
+    if s.state != RUNNING and p.pipeid != -1 :
+      call-c clib/delete_process_pipes(p.input, p.output, p.error)
```
CuppoJava: Check return code. Throw exception with linux-error if error.

runtime/driver.c
```diff
@@ -835,15 +836,22 @@ void initialize_launcher_process (){
...
-                    int input, int output, int error,
+                    int input, int output, int error, int pipeid,
                     Process* process){
   //Initialize launcher if necessary
   initialize_launcher_process();
   
   //Figure out unique pipe name
   char pipe_name[80];
-  sprintf(pipe_name, "/tmp/stanza_exec_pipe%ld", (long)getpid());
+  sprintf(pipe_name, "/tmp/stanza_exec_pipe_%ld_%ld", (long)getpid(), (long)pipeid);
```
CuppoJava: Delete the FIFOs on cleanup as well.

CuppoJava (changes requested):
Please fix the issues mentioned.

runtime/driver.c
jackbackrack: but this is /tmp.  i've closed the fifos just not deleted the files.  I've never seen anyone remove files from /tmp after being done.   are you sure?  that's why you use the PID so as to avoid any collisions.

jackbackrack: done.

PR #72, opened 2020-05-07
new rev of autodoc

jackbackrack:
add more options such as filtering on packages that are included, renaming of names for TOC, filtering on packages for which figures are generated, and tracing of calls to scripts.  scripts are now used for these operations with pkg name and filename passed as command line arguments and with the resulting string printed to standard output being the result (either true/false for filters or string for renaming).  the TOC is now generated as a hierarchy based on the TOC "pathname".

### DIFF
--- a/compiler/stz-auto-doc.stanza
+++ b/compiler/stz-auto-doc.stanza
@@ -385,7 +385,21 @@ defn parse-items (string:String, tag:String) -> Vector<ParsedItem> :
 
 val script-counter = to-seq(0 to false)
 
-;print definition into package file, printing anchors, and recording descriptions for later
+defn gen-fig (item:ParsedForm, pkg:IPackage, figs-dir:String, run-fig-generation?:True|False) -> String :
+  val id = next(script-counter)
+  val script-name = to-string("script%_.stanza" % [id])
+  val forms = unwrap-all(form(item))
+  val app    = forms[0]
+  val suffix = forms[1]
+  val code   = tailn(forms, 2)
+  val filename = to-string("%_/fig-%_.%_" % [name(pkg), id, suffix])
+  if run-fig-generation? :
+    val pathname = to-string("%_/%_" % [figs-dir, filename])
+    val output = to-string("call-system(\"%_\", [\"%_\", \"%_\", \"%_\", \\<s>%_<s>])" % [app, app, pathname, suffix, code])
+    spit(script-name, output)
+    call-system("stanza", ["stanza", "run", script-name])
+  to-string("figs/%_" % [filename])
+
 defn print-definition (o:OutputStream, nm:NameMap, e:IExp, descriptions:IntTable<String>, pkg:IPackage,
                        figs-dir:String, run-fig-generation?:True|False) :
   match(e) :
@@ -421,19 +435,7 @@ defn print-definition (o:OutputStream, nm:NameMap, e:IExp, descriptions:IntTable
           (item:ParsedText) :
             print(o, string(item))
           (item:ParsedForm) :
-            val id = next(script-counter)
-            val script-name = to-string("script%_.stanza" % [id])
-            val forms = unwrap-all(form(item))
-            val app    = forms[0]
-            val suffix = forms[1]
-            val code   = tailn(forms, 2)
-            val filename = to-string("%_/fig-%_.%_" % [name(pkg), id, suffix])
-            if run-fig-generation? :
-              val pathname = to-string("%_/%_" % [figs-dir, filename])
-              val output = to-string("call-system(\"%_\", [\"%_\", \"%_\", \"%_\", \\<s>%_<s>])" % [app, app, pathname, suffix, code])
-              spit(script-name, output)
-              call-system("stanza", ["stanza", "run", script-name])
-            val url = to-string("figs/%_" % [filename])
+            val url = gen-fig(item, pkg, figs-dir, run-fig-generation?)
             print(o, "\"%_\"" % [url])
     (e) :
       false
--- a/compiler/stz-auto-doc.stanza
+++ b/compiler/stz-auto-doc.stanza
@@ -97,27 +97,40 @@ TODO:
      this will need to be revisited so that text can be updated without rerunning all the generation programs.
 ;============================================================================== <DOC>
 
+public defstruct AutoDocPluginOutputException <: Exception :
+  name: String
+
+defmethod print (o:OutputStream, e:AutoDocPluginOutputException) :
+  print(o, "Unable to read tokens from plugin %_" % [name(e)])
+
 defn call-plugin (name:String, arguments:Seqable<String>) -> String :
   val proc = Process(name, cat([name], arguments), PROCESS-IN, PROCESS-OUT, PROCESS-ERR)
   val tokens = read-line(output-stream(proc))
   val response =
     match(tokens:List<Token>) :
       if empty?(tokens) :
-        fatal("Unable to read tokens from plugin %_" % [name])
+        throw(AutoDocPluginOutputException(name))
       else :
         unwrap-token(head(tokens))
     else :
-      fatal("Unable to read tokens from plugin %_" % [name])
+      throw(AutoDocPluginOutputException(name))
   wait(proc)
   response
 
 defn call-string-plugin (name:String, args:Seqable<String>) -> String :
   call-plugin(name, args)
 
+public defstruct AutoDocPluginBooleanException <: Exception :
+  name: String
+  response: String
+
+defmethod print (o:OutputStream, e:AutoDocPluginBooleanException) :
+  print("Read invalid response %_ from boolean plugin %_" % [response(e), name(e)])
+
 defn call-boolean-plugin (name:String, args:Seqable<String>) -> True|False :
   val response = call-string-plugin(name, args)
   if not contains?(response, ["true", "false"]) :
-    fatal("Read invalid response %_ from boolean plugin %_" % [response, name])
+    throw(AutoDocPluginBooleanException(name, response))
   response == "true"
 
 defn call-string-plugin (app-name:False|String, pkg:IPackage, default:IPackage -> String, trace?:True|False) -> String :
--- a/compiler/stz-auto-doc.stanza
+++ b/compiler/stz-auto-doc.stanza
@@ -97,6 +97,7 @@ TODO:
      this will need to be revisited so that text can be updated without rerunning all the generation programs.
  o add timeout check for plugin execution
  o add support for special characters in package pathnames
+ o add proper delete directory command that handles windows and escaping etc
 ;============================================================================== <DOC>
 
 public defstruct AutoDocPluginOutputException <: Exception :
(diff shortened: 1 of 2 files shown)


compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+
+defn call-plugin (name:String, arguments:Seqable<String>) -> False|String :
+  val proc = Process(name, cat([name], arguments), PROCESS-IN, PROCESS-OUT, PROCESS-ERR)
+  ; println("WRITING HELLO")
+  ; println(input-stream(proc), "\"HELLO\"")
+  ; flush(input-stream(proc))
+  val tokens = read-line(output-stream(proc))
+  val response = match(tokens:List<Token>) :
+    unwrap-token(head(tokens)) when not empty?(tokens)
+  wait(proc)
```
CuppoJava: Add check for proper termination here. Throw an exception if not.

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+     this will need to be revisited so that text can be updated without rerunning all the generation programs.
+;============================================================================== <DOC>
+
+defn call-plugin (name:String, arguments:Seqable<String>) -> False|String :
+  val proc = Process(name, cat([name], arguments), PROCESS-IN, PROCESS-OUT, PROCESS-ERR)
+  ; println("WRITING HELLO")
+  ; println(input-stream(proc), "\"HELLO\"")
+  ; flush(input-stream(proc))
+  val tokens = read-line(output-stream(proc))
+  val response = match(tokens:List<Token>) :
```
CuppoJava: What should proper operation be when the reader returns false (indicating EOF)? Seems like improper operation, maybe throw an exception?

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+    unwrap-token(head(tokens)) when not empty?(tokens)
+  wait(proc)
+  response
+
+defn call-string-plugin (name:String, args:Seqable<String>) -> String :
+  to-string(call-plugin(name, args))
```
CuppoJava: Conversion of `String|False` to a `String` is suspicious here. What is the benefit of a string representation of `"false"`?

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+
+defn call-boolean-plugin (name:String, args:Seqable<String>) -> True|False :
+  val response = call-string-plugin(name, args)
+  response == "true"
```
CuppoJava, line 149: For boolean plugin, generate proper error message if the returned value is not either `"true"` or '"false"`.

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+
+defn call-string-plugin (app-name:False|String, pkg:IPackage, default:IPackage -> String, trace?:True|False) -> String :
+  val res =
+    match(app-name:String) :
+      call-string-plugin(app-name, [to-string(name(pkg)), filename(info(pkg) as FileInfo)])
+    else : default(pkg)
+  println("CALL %_ PKG-NAME %_ PKG-FILENAME %_ RES %_" % [app-name, name(pkg), filename(info(pkg) as FileInfo), res]) when trace?
```
CuppoJava, line 156: Remove debug printlns.

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+  res
+
+defn call-boolean-plugin (app-name:False|String, pkg:IPackage, default:IPackage -> True|False, trace?:True|False) -> True|False :
+  val res =
+    match(app-name:String) :
+      call-string-plugin(app-name, [to-string(name(pkg)), filename(info(pkg) as FileInfo)]) == "true"
+    else : default(pkg)
+  println("CALL %_ PKG-NAME %_ PKG-FILENAME %_ RES %_" % [app-name, name(pkg), filename(info(pkg) as FileInfo), res]) when trace?
```
CuppoJava, line 164: Remove debug printlns

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,162 @@ defpackage stz/auto-doc :
...
+  key: String
+  children: Tuple<StringTree>
+
+defmethod print (o:OutputStream, node:StringTree) :
+  print(o, "StringTree(%_, %_)" % [key(node), children(node)])
+
+;Turns list of pathname strings into a tree of strings split at slashes
+;  assumes input strings are sorted and trees maintain sorting
+defn organize-strings (strings:Seqable<String>) -> Tuple<StringTree> :
+  val split-strings = to-list $ for s in strings seq : to-list $ split(s, "/")
```
CuppoJava, line 182: Use `stz/parsed-path` for nuanced handling of parsing a path. E.g. escaped / characters, etc.

compiler/stz-auto-doc.stanza
```diff
@@ -255,8 +378,11 @@ defn parse-items (string:String, tag:String) -> Vector<ParsedItem> :
         add(items, ParsedText(string[start to false]))
   items
 
+val script-counter = to-seq(0 to false)
+
 ;print definition into package file, printing anchors, and recording descriptions for later
-defn print-definition (o:OutputStream, nm:NameMap, e:IExp, descriptions:IntTable<String>, pkg:IPackage, figs-dir:String) :
+defn print-definition (o:OutputStream, nm:NameMap, e:IExp, descriptions:IntTable<String>, pkg:IPackage,
```
CuppoJava, line 431: Function is doing many things. Analyze results into datastructure first, and then separate IO into individual functions.

compiler/stz-auto-doc.stanza
```diff
@@ -336,9 +463,35 @@ defn copy-figs (master-figs-dir:String, package-name:Symbol, info:False|FileInfo
...
+  [ks, vs]
+  
+
 ;produce mdbook doc for dir directory using namemap on all given ipackages
 ;outputs summary, package files, and index files
-public defn gen-doc (md-docs-dir:String, nm:NameMap, i-all-ipackages:Seqable<IPackage>) :
+public defn gen-doc (md-docs-dir:String, nm:NameMap, i-all-ipackages:Seqable<IPackage>, fig-filter-filename:False|String, toc-renamer-filename:False|String, trace?:True|False) :
   val all-ipackages = to-tuple $ i-all-ipackages
   val dir = string-join $ [md-docs-dir "/src"]
   call-system("sh", ["sh", "-c", string-join $ ["mkdir -p " dir]])
```
CuppoJava: `dir` needs to be properly escaped for the shell. See `escape-shell-command`.

compiler/stz-auto-doc.stanza
```diff
@@ -353,16 +506,26 @@ public defn gen-doc (md-docs-dir:String, nm:NameMap, i-all-ipackages:Seqable<IPa
...
+  val env = HashTable<Symbol,String>()
+  val pkg-strings = to-tuple $ for pkg in sorted-packages seq :
+    pkg => call-string-plugin(toc-renamer-filename, pkg, { to-string(name(_)) }, trace?)
+  val sorted-pkg-strings = to-tuple $ qsort(value, pkg-strings)
+  ; val toc-trees = organize-strings(seq({ to-string(name(_)) }, sorted-packages))
+  val [string-sorted-packages, sorted-strings] = unzip $ sorted-pkg-strings
+  val toc-trees = organize-strings(sorted-strings)
+  print-toc(so, 1, toc-trees, string-sorted-packages)
+  for ipackage in sorted-packages do :
+    println("IPACKAGE NAME %_ INFO %_" % [name(ipackage), info(ipackage)]) when trace?
```
CuppoJava, line 559: Remove debug printlns.

CuppoJava (commented):
Suggested changes are for a deeper engineering polish. Please correct what you think is appropriate for now, and accumulate the others in a TODO note.

compiler/stz-auto-doc.stanza
jackbackrack: those allow to trace the run using a trace? argument.

compiler/stz-auto-doc.stanza
jackbackrack: those allow to trace the run using a trace? argument.

compiler/stz-auto-doc.stanza
CuppoJava: Oh gotcha. Didn't scroll far enough to see the trace? guard.

compiler/stz-auto-doc.stanza
CuppoJava: Gotcha. Didn't scroll far enough to see the trace? argument.

compiler/stz-auto-doc.stanza
jackbackrack: there's no way to check for wait returning unless i were to give it a timeout and write my own wait with a given timeout.

CuppoJava: Wait can return a few statuses. Most of them are an indicator of abnormal termination, so those should be checked for. Can you add a TODO note for the timeout check?

compiler/stz-auto-doc.stanza
CuppoJava: Guarded by trace?. Okay.

compiler/stz-auto-doc.stanza
jackbackrack: done.

compiler/stz-auto-doc.stanza
jackbackrack: done.

compiler/stz-auto-doc.stanza
jackbackrack: done.

compiler/stz-auto-doc.stanza
jackbackrack: but this is a package name not a filename.  I'm not imagining such involved characters in it.  do you?

compiler/stz-auto-doc.stanza
CuppoJava: Okay. Don't fail with `fatal` in this case. Throw an exception instead. Fatal is used to indicate that the compiler is implemented incorrectly, and we've somehow reached an invalid state. An exception indicates that the compiler is implemented correctly, but we've hit an unexpected error.

compiler/stz-auto-doc.stanza
jackbackrack: didn't we have this conversation before and i tried that it doesn't work because this is for "sh" not for a raw call-system.

CuppoJava: Wasn't it the other way? If you were calling call-system directly you wouldn't have to escape it because it would be directly used as input into the argv array of the C main function. But because you are relying upon `sh` to parse the arguments for you, you have to encode it.

compiler/stz-auto-doc.stanza
CuppoJava: Oh my mistake. This is for package names not file paths? Can you add a comment that makes that clear? 

It's very rare, but I do allow escaped slashes in package names. Just add a TODO note to handle this carefully later. It's not important right now.

compiler/stz-auto-doc.stanza
jackbackrack: split up function but stopped short of building an intermediary datastructure.  is this good enough.  doing an intermediary datastructure is a fair amount of work and would double the code and not make it any more clear.  it would be more complex and definitely give it more options but i was trying to keep the thing simple until i knew what people want.  right now people can control the figs on a per package basis using a filter.

CuppoJava: That's totally fine. Thanks for the polish.

compiler/stz-auto-doc.stanza
CuppoJava: Thanks

compiler/stz-auto-doc.stanza
```diff
@@ -11,40 +12,194 @@ defpackage stz/auto-doc :
...
+
+defn call-string-plugin (name:String, args:Seqable<String>) -> String :
+  call-plugin(name, args)
+
+public defstruct AutoDocPluginBooleanException <: Exception :
+  name: String
+  response: String
+
+defmethod print (o:OutputStream, e:AutoDocPluginBooleanException) :
+  print("Read invalid response %_ from boolean plugin %_" % [response(e), name(e)])
```
CuppoJava: Print to OutputStream o, not to standard out.

PR #73, opened 2020-05-12
allow private in scope level expressions

jackbackrack:
changed read-iexp to allow private wrappers around scope-level expressions.

### DIFF
--- a/compiler/stz-il-ir.stanza
+++ b/compiler/stz-il-ir.stanza
@@ -578,17 +578,16 @@ public defn Begin (es:List<IExp>, info:False|FileInfo) :
 ;Flatten and remove private expressions
 ;Adds only a single level of flattening and assumes nested levels are already flattened
 public defn scope-flatten (es:List<IExp>) :
-  for e in es seq-append :
-    match(e:IBegin) :
-      for ex in exps(e) seq-append :
-        match(ex:IVisibility) :
-          if visibility(ex) is Private :
-            match(exp(ex)) :
-              (e:IBegin) : exps(e)
-              (e) : List(e)
-          else : List(ex)
-        else: List(ex)
-    else: List(e)
+   for e in es seq-append :
+      match(e) :
+         (e:IBegin) :
+           scope-flatten(exps(e))
+         (e:IVisibility) :
+           if visibility(e) is Private :
+             scope-flatten(List(exp(e)))
+           else :
+             List(e)
+         (e) : List(e)
 
 public defn ScopeBegin (es:List<IExp>, info:False|FileInfo) :
    val es* = scope-flatten(es)
--- a/compiler/stz-il-ir.stanza
+++ b/compiler/stz-il-ir.stanza
@@ -580,7 +580,7 @@ public defn Begin (es:List<IExp>, info:False|FileInfo) :
 public defn scope-flatten (es:List<IExp>) :
    for e in es seq-append :
       match(e) :
-         (e:IBegin) : exps(e)
+         (e:IBegin) : scope-flatten(exps(e))
          (e:IVisibility) :
            if visibility(e) is Private :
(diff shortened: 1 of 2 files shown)


compiler/stz-il-ir.stanza
```diff
@@ -565,11 +565,28 @@ public defn subexps (e:IExp) -> Collection<IExp> & Lengthable :
 public defn flatten (es:List<IExp>) :
    for e in es seq-append :
       match(e) :
-         (e:IBegin) : exps(e)
+         (e:IBegin) : flatten(exps(e))
```
CuppoJava: Remove this change, and add a comment to the function saying that it only flattens a single level of begins. 
The original read-iexp algorithm ensures that all Begins are flattened, so it is not necessary to recurse.

compiler/stz-il-ir.stanza
```diff
@@ -565,11 +565,28 @@ public defn subexps (e:IExp) -> Collection<IExp> & Lengthable :
...
    else : IBegin(es*, info)
+
+;Flatten and remove/check visibility expressions
+public defn scope-flatten (es:List<IExp>) :
+   for e in es seq-append :
+      match(e) :
+         (e:IBegin) : scope-flatten(exps(e))
+         (e:IVisibility) :
+           if visibility(e) is-not Private :
+             lang/check/error!(info(e), "Only Private visibility allowed in scope-level expressions.")
```
CuppoJava: Move this error check out of the scope-flatten function. You can assume that any Visibility annotations will be Private. Throw a fatal error otherwise.

compiler/stz-il-ir.stanza
```diff
@@ -565,11 +565,28 @@ public defn subexps (e:IExp) -> Collection<IExp> & Lengthable :
...
    val es* = flatten(es)
    if empty?(es*) : IBegin(es*, info)
    else if empty?(tail(es*)) : head(es*)
    else : IBegin(es*, info)
+
+;Flatten and remove/check visibility expressions
+public defn scope-flatten (es:List<IExp>) :
+   for e in es seq-append :
+      match(e) :
+         (e:IBegin) : scope-flatten(exps(e))
```
CuppoJava: Follow the same invariant as above. Assume that all Begins are properly flattened on creation.

compiler/stz-input.stanza
```diff
@@ -54,6 +57,10 @@ defn ls-name-args (e:IExp) -> [IExp, List<IExp>] :
       (e:ILSOfT) : [class(e), args(e)]
       (e) : [e, List()]
 
+;Scope-level bodies include:
+;  $defn, $defn*, $defmethod, $defmethod*, $fn, $fn*, $branch,
+;  $ls-letexp, $ls-block, $ls-let, $ls-branch,
+;  $ls-defn, $ls-extern-fn, $ls-defn*, $ls-defmethod, $ls-defmethod*,
```
CuppoJava, line 64: Thanks!

CuppoJava (changes requested):
Make the changes to preserve the well-formed Begin invariant. Thanks!

compiler/stz-il-ir.stanza
jackbackrack: unfortunately, removing this doesn't allow this to remove the private statements.

compiler/stz-il-ir.stanza
jackbackrack: done.

compiler/stz-il-ir.stanza
jackbackrack: done.

compiler/stz-il-ir.stanza
jackbackrack: fixed this in the latest version.

PR #74, opened 2020-05-15
more stringent response checking and more rigorous testing

jackbackrack:


CuppoJava (changes requested):
Don't use `reader` anymore to read from the plugin input stream. Instead, just read in the entire output to a single string. (See the implementation of `slurp` in core for an example.)

Then you can keep all the original logic which was defined on Strings. (Such as testing for `"true"` and `"false"`.) And you can remove the stuff involving `unwrap-token` etc.

CuppoJava (approved):
Looks good.

Issue #82, opened 2021-08-06
Iteration order over a Queue in collections


PhilippeFerreiraDeSousa:
The iteration order of a queue is the opposite of the conventional sense. Technically the theoretical definition of a queue does not impose anything or whether you should be able to iterate over a queue, but this is counter intuitive to iterate in the opposite order than popping. Queues are FIFO and it makes sense to iterate in the same order: If you add 1 then 2 then 3 to the queue, iterating should give you 1 then 2 then 3 (and leave all 3 elements in the queue of course).
I checked in python (deque from collections) and java (java.util.LinkedList implementation of java.util.Queue) and they do what I was expecting.

Stanza:
```
$ jstanza repl
stanza> val q = Queue<Int>()
stanza> add(q, 1)
stanza> add(q, 2)
stanza> add(q, 3)
stanza> println(to-tuple $ q)
[3 2 1]
```

Python:
```
lbstanza$ python3.8
Python 3.8.0 (default, Feb 25 2021, 22:10:10) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections import deque
>>> d = deque()
>>> d.append(1)
>>> d.append(2)
>>> d.append(3)
>>> d
deque([1, 2, 3])
>>> list(d)
[1, 2, 3]
```

Java:
```
import java.util.LinkedList;
import java.util.Queue;

class Main
{
	// Iterate through Queue in Java
	public static void main(String[] args)
	{
		Queue<Integer> queue = new LinkedList<Integer>();
		queue.add(1);
		queue.add(2);
		queue.add(3);

		for (Integer item: queue) {
			System.out.println(item);
		}
	}
}
```
gives
```
1
2
3
```

PR #83, opened 2021-08-06
[WIP] Add basic write helpers as well as printer/writer for each type in collections

PhilippeFerreiraDeSousa:
This PR enforces:
- `print` gives human readable output,
- `write` gives machine readable output.

It contains:
- basic write helpers
- printer / writer for each type in collections

3 things missing:
- need to talk with @CuppoJava if he is okay with this.
I think that this should be in collections as anybody that handles an hashtable should have its printer available and not risk to have forgotten to import another package and have to stop the program and add the import to display some variable content in your program.
- show other people the proposed `print` outputs for types in collections, if somebody has a better proposition
- I need to finish the test file (see TODO in it). But I get a Stack Overflow for no apparent reason:
```
lbstanza$ stanza run-test core/core.stanza tests/test-collections-printers.stanza 
FATAL ERROR: Stack overflow
Segmentation fault (core dumped)
```
even when commenting out all changes in core/core.stanza

2 remarks that would require breaking changes to change:
- queue iteration order: https://github.com/StanzaOrg/lbstanza/issues/82
- print is human readable (no theoretical constraint, could be anything, it is supposed to look good) while write is machine readable (there is a hard constraint: getting a valid stanza expression that evaluates to the same object without having to import any other package than the package containing the definition of the type and the package where the writer is defined is different). So a printer is not a writer but a writer could be a printer. That is why `write` calling `print` by default is wrong. On the contrary, `print` could call `write` by default.

### DIFF
--- a/core/collections.stanza
+++ b/core/collections.stanza
@@ -37,6 +37,15 @@ defstruct SetItem<K> :
 defmethod print (o:OutputStream, x:SetItem) :
   print(o, "(%_) %_" % [hash(x), key(x)])
 
+defn indented-list (items: Seqable) :
+  new Printable :
+    defmethod print (o:OutputStream, this) :
+      val o2 = IndentedStream(o)
+      do(lnprint{o2, _}, items)
+    defmethod write (o:OutputStream, this) :
+      val o2 = IndentedStream(o)
+      do(lnwrite{o2, _}, items)
+
 ;Binary search
 ;Returns n such that the first n numbers in xs < v
 defn bsearch<?T,?S> (less?: (T, S) -> True|False,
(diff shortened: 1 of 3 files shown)


core/collections.stanza
```diff
@@ -797,11 +838,27 @@ public defn to-hashtable<K,V> (es:Seqable<KeyValue<K,V>>) -> HashTable<K,V> :
     t[key(e)] = value(e)
   t
 
+public defn to-hashtable<?K,?V> (es:Seqable<KeyValue<?K,?V>>) -> HashTable<K,V> :
+  to-hashtable<K,V>(es)
```
PhilippeFerreiraDeSousa, line 842: I don't want to define those but i don't know how to print the types in the writer; `"to-hashtable<String, Int>(...)"` for example instead of `"to-hashtable(...)"`

PR #88, opened 2021-08-25
Design & implement elastic markable memory chunks

OlegPliss:
The heap is represented by two contiguous resizable memory chunks, one of which stores the data, the other stores a marking bit for each 64-bit aligned word of the data. The mark can be set, cleared and tested. `test-and-set` is an additional fused operation for frequently used sequence of test and set.

Marked words can be iterated in a given address range of the heap. The iteration is going to be used for testing & debugging, in recovery from marking stack overflow and in remembered set scan of young collections.  

The heap is going to be aggregated to VMState. It can be initialized, finalized and resized in place.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -1004,12 +1004,20 @@ lostanza defn collect-garbage (size:long) -> long :
 ;================== New Garbage Collector ===================
 ;============================================================
 
+;Define various constants for the relations between
+;bits, bytes, and longs.
 lostanza val LOG-BITS-IN-BYTE:long = 3
 lostanza val LOG-BYTES-IN-LONG:long = 3
 lostanza val LOG-BITS-IN-LONG:long = LOG-BYTES-IN-LONG + LOG-BITS-IN-BYTE
 lostanza val BYTES-IN-LONG:long = 1 << LOG-BYTES-IN-LONG
 lostanza val BITS-IN-LONG:long = 1 << LOG-BITS-IN-LONG
 
+;Structure for representing a Heap space.
+;- start is the starting address of the heap.
+;- top is the top address of the heap.
+;- limit is equal to start + number of currently available bytes in the heap.
+;- size is the maximum size that the heap can be expanded to.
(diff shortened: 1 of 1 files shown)


core/core.stanza
```diff
@@ -999,6 +1000,193 @@ lostanza defn collect-garbage (size:long) -> long :
...
+  val index = bit-index(heap, p)
+  val address = bit-address(heap, index)
+  [address] = [address] & (~ bit-mask(index))
+  return 0
+
+lostanza defn clear-mark (heap:ptr<Heap>, start:ptr<?>, limit:ptr<?>) -> int :
+  check-range(heap, start, limit)
+
+  val start-bit-index = bit-index(heap, start)
+  val end-bit-index = bit-index(heap, limit - 1)
```
CuppoJava, line 1187: This doesn't actually make any difference, but would the following be a more accurate representation of your intent? This is based upon my understanding of what `bit-index` does.

```
val end-bit-index = bit-index(heap, limit - BYTES-IN-LONG)
```

core/core.stanza
```diff
@@ -999,6 +1000,193 @@ lostanza defn collect-garbage (size:long) -> long :
...
+    if (bits & 0xFFL) == 0 :
+      bits = bits >> 8
+      p = p + (8 * BYTES-IN-LONG)
+    if (bits & 0xFL) == 0 :
+      bits = bits >> 4
+      p = p + (4 * BYTES-IN-LONG)
+    if (bits & 0x3) == 0 :
+      bits = bits >> 2
+      p = p + (2 * BYTES-IN-LONG)
+    if (bits & 1) != 0 :
```
CuppoJava, line 1154: Does this condition always return true? Based upon the logic in line 1137 and the shifts in the previous lines.

core/core.stanza
OlegPliss: I expected this question. Yes, the result is the same but 1 is shorter than BYTES-IN-LONG.

CuppoJava (approved):
Looks good to me!

PR #89, opened 2021-08-31
Build script improvements for cross-compilation

tylanphear:
This patch improves the existing build script (`scripts/make.sh`) to make cross-compilation easier. It also streamlines some aspects of the current build process.

The main change is the addition of a `<platform>` parameter to `scripts/make.sh`. This can be one of `windows`, `linux`, or `os-x`, corresponding (I think) to the values accepted by the Stanza compiler with `-platform`. If one is not passed, the script errors out (could detect current platform in the future?)

Additionally, this patch makes it so we only build the platform requested (previously would always build OS-X and Linux). It also adds a finish script for Windows (since one did not previously exist).

### DIFF
--- a/scripts/make.sh
+++ b/scripts/make.sh
@@ -16,6 +16,24 @@ fi
 
 STANZA="$1"
 PLATFORM="$2"
+COMMAND="${3:-compile-clean}" # default to `compile-clean` if no command is passed
+
+case "$COMMAND" in
+    compile)
+        CLEAN=0 COMPILE=1 FINISH=1 ;;
+    compile-clean)
+        CLEAN=1 COMPILE=1 FINISH=1 ;;
+    compile-without-finish)
+        CLEAN=0 COMPILE=1 FINISH=0 ;;
+    compile-clean-without-finish)
+        CLEAN=1 COMPILE=1 FINISH=0 ;;
+    clean)
+        CLEAN=1 COMPILE=0 FINISH=0 ;;
+    *) cat 1>&2 <<EOF
+Error: unsupported/unrecognized command: \`$COMMAND\`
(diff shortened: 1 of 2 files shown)


CuppoJava (changes requested):
Thanks! The nature of the changes look good, but is it possible to write them with simpler bash constructs?

tylanphear:
> Thanks! The nature of the changes look good, but is it possible to write them with simpler bash constructs?

I had a feeling that might be the case (bad habit). I just pushed a cleaner version. Let me know if it's better.

CuppoJava (changes requested):
The simpler style is much better, thank you. 

Can you add one knob for me? I don't currently have the cross-compiler set up, so I would like to run all of the steps except for the very last call to the `xfinish.sh` scripts.

Can you provide some argument that controls whether or not that last step is executed?

tylanphear:
Pushed a change that adds a third (optional) parameter to `scripts/make.sh` which serves as a command to the build script. This allows you to only run the parts of the script that you want.

Commands added and their behaviors are as follows:
* `compile`:
   - compile Stanza pkgs & compiler
   - finish executable
* `compile-without-finish`:
   - compile Stanza pkgs & compiler
* `compile-clean-without-finish`:
    - clean Stanza pkgs & build artifacts
    - compile Stanza pkgs & compiler
* `compile-clean`:
    - clean Stanza pkgs & build artifacts
    - compile Stanza pkgs & compiler
    - finish executable
* `clean`:
    - clean Stanza pkgs & build artifacts

If no command is passed, the script defaults to `compile-clean` (perform all steps -- which was the previous behavior).

Let me know if you think this is overkill, if the names could be improved, or if there is some combination missing you'd like me to add.

CuppoJava (approved):
Looks good. Thank you for adding the additional control.

PR #90, opened 2021-08-31
Added comments during reading.

CuppoJava:
Added comments during code read through, and made minor modifications for style. Let me know if I have misunderstandings.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -1004,12 +1004,20 @@ lostanza defn collect-garbage (size:long) -> long :
 ;================== New Garbage Collector ===================
 ;============================================================
 
+;Define various constants for the relations between
+;bits, bytes, and longs.
 lostanza val LOG-BITS-IN-BYTE:long = 3
 lostanza val LOG-BYTES-IN-LONG:long = 3
 lostanza val LOG-BITS-IN-LONG:long = LOG-BYTES-IN-LONG + LOG-BITS-IN-BYTE
 lostanza val BYTES-IN-LONG:long = 1 << LOG-BYTES-IN-LONG
 lostanza val BITS-IN-LONG:long = 1 << LOG-BITS-IN-LONG
 
+;Structure for representing a Heap space.
+;- start is the starting address of the heap.
+;- top is the top address of the heap.
+;- limit is equal to start + number of currently available bytes in the heap.
+;- size is the maximum size that the heap can be expanded to.
+;- bitset is the starting address of the marking bits for the heap.
 lostanza deftype Heap :
   var top: ptr<long>
   var limit: ptr<long>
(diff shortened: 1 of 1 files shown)


core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
 lostanza defn clear (start:ptr<?>, size:long) -> ptr<?> :
   return call-c clib/memset(start, 0, size)
 
-lostanza defn initialize-heap (heap:ptr<Heap>, min-size:long, max-size:long) -> int :
-  ;Assert 0 <= min-size && min-size <= max-size
+;Initialize the current heap structure.
+;- min-size is the initial size of the heap.
+;- max-size is the maximum size that heap can be expanded to.
+lostanza defn initialize-heap (heap:ptr<Heap>, min-size:long, max-size:long) -> ref<False> :
+  ;Precondition: Assert 0 <= min-size && min-size <= max-size
```
CuppoJava, line 1048: Added a precondition. Will this work?

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
+;Returns 'start'. 
 lostanza defn clear (start:ptr<?>, size:long) -> ptr<?> :
   return call-c clib/memset(start, 0, size)
 
-lostanza defn initialize-heap (heap:ptr<Heap>, min-size:long, max-size:long) -> int :
-  ;Assert 0 <= min-size && min-size <= max-size
+;Initialize the current heap structure.
+;- min-size is the initial size of the heap.
+;- max-size is the maximum size that heap can be expanded to.
+lostanza defn initialize-heap (heap:ptr<Heap>, min-size:long, max-size:long) -> ref<False> :
```
CuppoJava, line 1047: A Stanza convention that I'm trying to move to is to return `ref<False>` when the return value is not meaningful. This has been changed systematically everywhere.

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
+  ;No meaningful return value.
+  return false
 
-lostanza defn expand-heap (heap:ptr<Heap>, size:long) -> int :
+;Ensure that the given heap is at least the given size.
+;Assumes that the given size is less than the maximum size that heap can be expanded to.
+lostanza defn expand-heap (heap:ptr<Heap>, size:long) -> ref<False> :
+  ;Precondition: Ensure that size is less than maximum allowed size.
+  #if-not-defined(OPTIMIZE) :
+    if heap.size < size.size :
```
CuppoJava: Is this an appropriate precondition?

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
     clear(heap.bitset + current-bitset-size, desired-bitset-size - current-bitset-size)
-  return 0
+    
+  ;No meaningful return value.
+  return false
 
-lostanza defn check-range (heap:ptr<Heap>, p:ptr<?>) -> int :
+;Sanity check: Ensure that the given pointer points to within the given heap.
+;Calls fatal if it is not.
+lostanza defn ensure-pointer-in-heap! (heap:ptr<Heap>, p:ptr<?>) -> ref<False> :
```
CuppoJava, line 1108: Renamed this function to aid my reading.

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
+    if p < heap.start or  p >= heap.top :
       fatal("Pointer is outside of heap.")
-  return 0
+  return false
 
-lostanza defn check-range (heap:ptr<Heap>, start:ptr<?>, limit:ptr<?>) -> int :
+;Sanity check: Ensure that the given address range: start (inclusive) to limit (exclusive)
+;is contained within the given heap.
+;Calls fatal if it is not.
+lostanza defn ensure-address-range-in-heap! (heap:ptr<Heap>, start:ptr<?>, limit:ptr<?>) -> ref<False> :
```
CuppoJava, line 1117: Renamed this function to aid my reading.

core/core.stanza
```diff
@@ -1169,23 +1262,35 @@ lostanza defn iterate-marked (heap:ptr<Heap>, start:ptr<?>, limit:ptr<?>, f:ptr<
...
-  while bit-address < end-bit-address :
-    if [bit-address] != 0 :     ;Rarely taken
-      iterate-bit-word(heap, p, [bit-address], f)
-    bit-address = bit-address + BYTES-IN-LONG
-    p = p + (BITS-IN-LONG * BYTES-IN-LONG)
 
-  ;Process the last bit word
-  iterate-bit-word(heap, p, [end-bit-address] & end-bit-mask, f)
-  return 0
+  else :
```
CuppoJava, line 1273: Stylistically, I like the `else` branch to signal that this is a different case.

CuppoJava (commented):
I added comments on all non-comment modifications.

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
+;- size is the maximum size that the heap can be expanded to.
+;- bitset is the starting address of the remembered set for the heap.
 lostanza deftype Heap :
   var top: ptr<long>
   var limit: ptr<long>
   var start: ptr<long>
   var size: long
   var bitset: ptr<long>
 
+;Given the current number of bytes in the heap, return
```
OlegPliss, line 1028: bytes --> longs

core/core.stanza
OlegPliss: Ok. I would prefer a distinct type for that.

core/core.stanza
OlegPliss: Sure :) There are other loose ends in the code too.

core/core.stanza
OlegPliss: `heap.size < size`

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
+      fatal("Address range is outside of heap.")
+  return false
 
+;Return the index of the remembered bit that acts as the mark for the given heap pointer.
 lostanza defn bit-index (heap:ptr<Heap>, p:ptr<?>) -> long :
-  check-range(heap, p)
+  ensure-pointer-in-heap!(heap, p)
   return (p - heap.start) >> LOG-BYTES-IN-LONG
 
+;Returns the pointer in the remembered set containing the given mark bit.
```
OlegPliss: I would not use the term `remembered set` here. Bitset of the heap provides the marking capability for individual heap cells / longs.  Both reachable objects and references from old gen to young gen can be marked.

core/core.stanza
```diff
@@ -1004,98 +1004,154 @@ lostanza defn collect-garbage (size:long) -> long :
...
 lostanza defn bit-address (heap:ptr<Heap>, bit-index:long) -> ptr<long> :
-  return heap.bitset + ((bit-index >> LOG-BITS-IN-LONG) << LOG-BYTES-IN-LONG)
+  return heap.bitset + (bit-index >> LOG-BITS-IN-LONG) << LOG-BYTES-IN-LONG
 
+;Returns bit-index % 64.
```
OlegPliss: I would better use BITS-IN-LONG here. I am not sure if `BITS-IN-LONG` is a right name. If lostanza had typedefs, I would define `heap-cell` as `long` and call the constant `BITS-IN-HEAP-CELL`.

core/core.stanza
```diff
@@ -1123,43 +1189,70 @@ lostanza defn clear-mark (heap:ptr<Heap>, start:ptr<?>, limit:ptr<?>) -> int :
...
+;and the 64-bit mask given in 'bit-word', 
+;call 'f' on every pointer corresponding to a 1 in 'bit-word'. 
+lostanza defn iterate-bit-word (heap:ptr<Heap>, start:ptr<?>, bit-word:long,
+                                f:ptr<((ptr<Heap>, ptr<?>) -> ref<False>)>) -> ref<False> :
   ;Ones are rare, expect zeroes
   var p:ptr<?> = start
   var bits:long = bit-word
   while bits != 0 :
     ;TODO: introduce BSF instruction and use it here
+    ;If the next 32 bits in 'bits' is zero, then
```
OlegPliss, line 1216: `BSF` instruction will be introduced soon. It will render these bit masks, shifts and comments obsolete.

core/core.stanza
OlegPliss: Early returns are strongly preferred by many coding standards, i.e. LLVM. They reduced visual complexity of the code.

Use Early Exits and continue to Simplify Code
https://llvm.org/docs/CodingStandards.html

OlegPliss (commented):
I prefer early return over `else` branch.

core/core.stanza
CuppoJava: Oh! This function returns the number of longs in the bitset! 

It seems that the other places in the code is treating the result as the number of bytes. Is that a bug?

e.g. 
```
  val min-heap-size = round-up-to-whole-pages(min-size)
  val max-heap-size = round-up-to-whole-pages(max-size)
```
and
```
  clear(heap.bitset, min-bitset-size)
```

core/core.stanza
CuppoJava: Yes, I agree with you. That is part of a larger change, that I'll leave for later.

core/core.stanza
CuppoJava: Thanks!

core/core.stanza
CuppoJava: Thanks! I've updated the comments to reflect that.

core/core.stanza
CuppoJava: Okay!

core/core.stanza
CuppoJava: From a functional programming background, I would disagree with that standard. 

The nesting level is proportional to the number of conditions that currently apply to the code that I am reading. I think the early returns make that harder, because I have to go back and recover what are the assumptions that hold for the code block I'm reading.

OlegPliss: Then you will have to correct my code all the time :) I really feel early exits make the code simpler and code in this style for many years.

core/core.stanza
OlegPliss: It is in bytes but rounded to longs.

core/core.stanza
CuppoJava: Okay :). I'm not fussy about this, though I've made my personal preference known.

core/core.stanza
CuppoJava: Can you verify whether this is the intended behaviour?
```
lostanza let :
  val x = 64L * 8L
  val result = bitset-size(x)
  call-c clib/printf("bitset-size(%ld) = %ld\n", x, result)
```
prints
```
bitset-size(512) = 1
```

But I expect it to print 8.

A heap size of 64 * 8 bytes holds 64 pointers.
64 pointers requires 64 bits for marking.
64 bits for marking corresponds to 8 bytes.

PR #91, opened 2021-08-31
Introduce `stz_*` types in Stanza runtime

tylanphear:
In the Stanza runtime, we have a bunch of C code that is called from Stanza. In doing so, it is important that we make sure that the size/signedness of types in Stanza is preserved when passed over the boundary to C. This is especially necessary because, while types in Stanza may often have the same name as types in C, their representation can be very different.

One place where our usage of C types caused issue was `void* stz_malloc(long)`, which has the corresponding Stanza definition `extern stz_malloc: long -> ptr<?>`. In Stanza, `long` is always an signed 64 bit integer, whereas in C `long` can be any integer at least 32 bits wide. Indeed, in C on Windows `long` is a 32-bit signed integer and on Linux `long` is a 64-bit signed integer.

In order to control the representation of types passed along the boundary, this patch introduces `stz_*` types (e.g. `stz_long`) whose representation matches exactly that of the corresponding type in Stanza. These types should be used at the boundary between C and Stanza in order to prevent size/signedness issues. For example, the above `stz_malloc` becomes `void* stz_malloc(stz_long)`. Additionally included are also a set of macros for conversion between Stanza strings (`stz_byte*`) and C-strings (`char*`).

### DIFF
--- a/runtime/driver.c
+++ b/runtime/driver.c
@@ -18,45 +18,46 @@
 #include<dirent.h>
 #include<pthread.h>
 
+#include "types.h"
+
 //       Forward Declarations
 //       ====================
 
 //FMalloc Debugging
 void init_fmalloc ();
 
-//Stanza Alloc
-void* stz_malloc (long size);
+void* stz_malloc (stz_long size);
 void stz_free (void* ptr);
 
 //     Stanza Defined Entities
 //     =======================
 typedef struct{
-  char* heap;
-  char* heap_top;
-  char* heap_limit;
-  char* free;
-  char* free_limit;
-  uint64_t current_stack;  
-  uint64_t system_stack;  
+  stz_byte* heap;
+  stz_byte* heap_top;
+  stz_byte* heap_limit;
+  stz_byte* free;
+  stz_byte* free_limit;
+  stz_long current_stack;
+  stz_long system_stack;
 } VMInit;
 
 typedef struct{
-  uint64_t returnpc;
-  uint64_t liveness_map;
-  uint64_t slots[];
+  stz_long returnpc;
+  stz_long liveness_map;
+  stz_long slots[];
 } StackFrame;
 
 typedef struct{
-  int pool_index;
-  int mark;
+  stz_int pool_index;
+  stz_int mark;
   StackFrame frames[];
 } StackFrameHeader;
 
 typedef struct{
-  uint64_t size;
+  stz_long size;
   StackFrame* frames;
   StackFrame* stack_pointer;
-  int pc;
+  stz_long pc;
 } Stack;
 
 //     Macro Readers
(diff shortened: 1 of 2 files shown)


OlegPliss:
There are other places where lostanza`long` is used instead of C `size_t`, i.e. `memset` in core.stanza.

tylanphear:
> There are other places where lostanza`long` is used instead of C `size_t`, i.e. `memset` in core.stanza.

Yes, this is true. At least, since we only compile on 64-bit, we know that `sizeof(long) == sizeof(size_t)`, but there is still a signedness mismatch there, since `long` is signed and `size_t` is unsigned.

I think the correct solution to this is to create types in the Stanza `clib` that correspond to C types. I know that this is the approach used in the `libc` crate in Rust and the `ctypes` module in Python. The idea is trivial to state, but the implementation could be more involved, as I don't know of any ways to control representation in Stanza (outside of the existing builtin types).

OlegPliss:
Usually the solution for native function interface is two-fold: introduce native types to the language (i.e. `size_t` in stanza clb) and language types to native code (i.e. `stz_long` in `stanza-types.h` included to `runtime/driver.c`).

tylanphear:
I agree 100%. With this PR we will have the latter half of the solution. We will almost certainly need the former in the future.

runtime/driver.c
```diff
@@ -18,165 +18,165 @@
...
-int file_set_length (FILE* f, int64_t size) {
-  return ftruncate(fileno(f), size);
+stz_int file_set_length (FILE* f, stz_long size) {
+  return (stz_int)ftruncate(fileno(f), size);
 }
 
-int64_t file_read_block (FILE* f, char* data, int64_t len) {
-  return fread(data, 1, len, f);
+stz_long file_read_block (FILE* f, char* data, stz_long len) {
+  return (stz_int)fread(data, 1, len, f);
```
CuppoJava: Should this be `stz_long` since the return type is `stz_long`?

runtime/driver.c
```diff
@@ -18,165 +18,165 @@
...
 }
 
-int64_t file_write_block (FILE* f, char* data, int64_t len) {
-  return fwrite(data, 1, len, f);
+stz_long file_write_block (FILE* f, char* data, stz_long len) {
+  return (stz_int)fwrite(data, 1, len, f);
```
CuppoJava: Should this be stz_long since the return type is stz_long?

runtime/driver.c
```diff
@@ -18,165 +18,165 @@
...
   StackFrame frames[];
 } StackFrameHeader;
 
 typedef struct{
-  uint64_t size;
+  stz_long size;
   StackFrame* frames;
   StackFrame* stack_pointer;
-  int pc;
+  stz_long pc;
```
CuppoJava, line 60: Thank you for catching that.

CuppoJava (changes requested):
Thank you! Small question about `fread` et al.

runtime/driver.c
tylanphear: Yes, absolutely. Not sure what happened here.

runtime/driver.c
tylanphear: Ditto to above

tylanphear:
> Thank you! Small question about `fread` et al.

Should be fixed now! Let me know if you have further comments/concerns.

CuppoJava (approved):
Looks good.

PR #92, opened 2021-08-31
Allowed single file input in `definitions-database`

m-hilgendorf:
This PR extends the `definitions-database`: 

- Allow the `-verbose` flag to be passed
- Allows individual `.stanza` files as inputs

This allows the command to index definitions for source files as needed.

### DIFF
--- a/compiler/stz-defs-db.stanza
+++ b/compiler/stz-defs-db.stanza
@@ -21,7 +21,7 @@ defpackage stz/defs-db :
 ;============================================================
 
 public defstruct DefsDbInput :
-  proj-files: Tuple<String>
+  files: Tuple<String>
   platform: Symbol|False
   flags: Tuple<Symbol>
   optimize?: True|False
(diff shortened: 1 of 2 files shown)


CuppoJava (commented):
Thank you for the PR. Before I can merge it, I need some more context about how it is supposed to be used and what it is supposed to calculate.

What is the following call supposed to do: 
```
stanza definitions-database myfile.stanza -o mydb
```

And what is the following call supposed to do:
```
stanza definitions-database myproj.proj myfile.stanza -o mydb
```

m-hilgendorf:
These calls are analogous to the following:

```
stanza compile myfile.stanza -o myprogram
stanza compile myproj.proj myfile.stanza -o myprogram
```

If `myfile.stanza` is well formed, its definitions should be able to be compiled, just as if it should be able to be compiled as a program or run in the VM. 

The ultimate goal is to be able to compile a table of definitions within a single file without compiling all of its dependencies.

PR #93, opened 2021-09-01
Windows runtime support

tylanphear:
This patch implements/improves Stanza runtime support for Windows. Previously the runtime would not build under Windows (due to changes made in the interim), however with these changes applied I have been able to successfully build the compiler (on Linux using cross-compilation) and then build the compiler with itself.

Things implemented in this PR:
* Windows process library
* Windows linenoise library (stubbed)
* Windows resizeable memory regions support
* Windows path support (not necessarily comprehensive)

### DIFF
--- a/runtime/driver.c
+++ b/runtime/driver.c
@@ -1,7 +1,8 @@
 #ifdef PLATFORM_WINDOWS
-  #include<Windows.h>
+  #include<windows.h>
 #else
   #include<sys/wait.h>
+  #include<sys/mman.h>
 #endif
 #include<stdint.h>
 #include<unistd.h>
@@ -14,7 +15,6 @@
 #include<string.h>
 #include<sys/stat.h>
 #include<sys/types.h>
-#include<sys/mman.h>
 #include<dirent.h>
 #include<pthread.h>
 
--- a/runtime/driver.c
+++ b/runtime/driver.c
@@ -29,6 +29,11 @@
 static void init_fmalloc ();
 #endif
 
+static void exit_with_error (){
+  fprintf(stderr, "%s\n", strerror(errno));
+  exit(-1);
+}
+
 //     Stanza Defined Entities
 //     =======================
 typedef struct{
@@ -487,11 +492,6 @@ typedef struct {
 //-------------------- Utilities -----------------------------
 //------------------------------------------------------------
 
-static void exit_with_error (){
-  fprintf(stderr, "%s\n", strerror(errno));
-  exit(-1);
-}
-
 static int count_non_null (void** xs){
   int n=0;
   while(xs[n] != NULL)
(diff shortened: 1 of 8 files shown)


runtime/driver.c
```diff
@@ -473,122 +493,163 @@ typedef struct {
...
 //Allocates a segment of memory that is min_size allocated, and can be
-//resized up to max_size. 
-void* stz_memory_map (long min_size, long max_size) {
-  void* p = mmap(NULL, max_size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+//resized up to max_size.
+void* stz_memory_map (stz_long min_size, stz_long max_size) {
+  void *p;
+
+#ifdef PLATFORM_WINDOWS
+  p = VirtualAlloc(NULL, (SIZE_T)max_size, MEM_COMMIT | MEM_RESERVE, STZ_MPROTECT_NONE);
```
OlegPliss: This code reserves and commits max_size memory.  We want to reserve max_size but to commit min_size.

tylanphear: Yeah, I had a feeling this was wrong. Originally I only had `MEM_RESERVE` but the call was failing for some reason (hence I added `MEM_COMMIT` and it went away.)

I'm not sure, but it may have been due to an unrelated size issue (`SIZE_T` vs `DWORD`) that I've since fixed. I'll investigate a bit further and see about fixing this.

OlegPliss: If you are not comfortable with this part, you can create JIRA task in Stanza project and assign it to me.

CuppoJava:
Hi Tyler,

Thank you for the progress you've made. Can you split out the following two features so that we can review/edit/merge them separately?

1) The windows process runtime.
- I would like to move more logic out of C and into Stanza/LoStanza, and handle the compatibility #if cases on the Stanza side. 
- Please use the Stanza `escape-shell-command` function to merge the call arguments into a single string. It handles some additional edge cases. I believe this will eliminate `allocating_sprintf`, `create_command_line_from_argv`.
- Please use the Stanza `norm-path` command to convert the filepaths into the appropriate Windows path when creating the Process. This will eliminate `normalize_path`. 
- Windows has a version of `CreateProcess` called `CreateProcessA` that handles the logic of searching through the PATH (and other system directories) for the appropriate executable to call. (https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa). Use that instead of using our own implementation. This should eliminate `is_executable_path`, `contains_path_separator`, `find_candidate_in_path`.
- Let's handle the #if compatibility cases on the Stanza side instead of on the C side. So this will eliminate the `delete_process_pipes` and `initialize_launcher_process` stubs.

2) The windows implementation of `stz_memory_map`, et al. 
- The linux versions of these functions, written by Oleg, already assume that the input arguments are multiples of the system page size so please simplify the code by making the same assumptions.
- Can you explain a little about why the Windows code is so much longer than the linux code? Are there more nuances to the Windows API?

tylanphear:
Yes, I can easily split the PR into two (since the memory map stuff is mostly self-contained), however there are still some things (cleaning up headers) that are common to both.

Going through your points one by one:

1. Windows Process Runtime

  * Yes, agreed.
  * The interface of `launch_process` accepts a `stz_byte*` (stanza str) for the executable and a `stz_byte**` (array of stanza strs) for the arguments. I see that `escape-shell-command` turns a list of arguments into a single command line -- would the idea be to pass `NULL` for the arguments and just pass the entire command line for the executable?
  * I am fine with this, however one note: `norm-path` seems to only handle the slash conversion, whereas `normalize_path` uses `GetFullPathName` to not only do that, but also to expand relative paths. I'm assuming we're fine with that (not sure what the behavior is on Linux)
  * `CreateProcess` is actually a macro that expands to `CreateProcessA` (or `CreateProcessW`) depending on the value of `UNICODE`, so we are actually using `CreateProcessA`. The PATH-searching behavior is only triggered when `CreateProcess` is called with a `NULL` executable (right now we are passing one directly). Then Windows will perform a lookup based on the first argument of the command line.
  * Yes, agreed.

I think a simple way to address multiple of your requests is to adjust the interface of `launch_process` on Windows to only take a command line (and to eliminate other unused arguments). Then we can just pass that to `CreateProcess` and have the Windows API do the heavy lifting for us.

2. Windows implementation of `stz_memory_map`

  * I had a suspicion that this was only an issue in testing (where I wasn't being careful to adhere to that restriction). Agreed.
  * Yes, there is one platform-specific detail that complicates things. On Linux, when you request a memory mapping, the operating system reserves pages and commits them to memory automatically as-needed. On Windows, however, this commitment process must be done manually. Hence when we allocate our resizable segment, we `MEM_RESERVE` the entire maximum size, then `MEM_COMMIT` the minimum size (the initial size). As we resize, we have to take care to `MEM_COMMIT` and `MEM_DECOMMIT` as we grow/shrink respectively. This way the semantics of the POSIX implementation are preserved.

runtime/driver.c
```diff
@@ -505,49 +540,144 @@ static int make_pipe (char* prefix, char* suffix){
...
+  if (p == NULL) {
+    return;
+  }
+
+  if (!VirtualFree(p, 0, MEM_RELEASE)) {
+    exit_with_error();
+  }
+}
+
+#define ROUND_UP(n, divisor) ((((n) + (divisor) - 1) / (divisor)) * (divisor))
```
OlegPliss, line 577: It is not necessary to round up to whole pages here. It is already done at stanza side.

runtime/driver.c
```diff
@@ -505,49 +540,144 @@ static int make_pipe (char* prefix, char* suffix){
...
+#define PAGE_SIZE 0x1000
+#define ROUND_UP_TO_NEAREST_PAGE(n) ROUND_UP((n), PAGE_SIZE)
+
+//Resizes the given segment.
+//old_size is assumed to be the size that is already allocated.
+//new_size is the size that we desired to be allocated.
+void stz_memory_resize (void* p, stz_long old_size, stz_long new_size) {
+  stz_long min_size, max_size;
+  enum {
```
OlegPliss, line 586: This enum is not necessary. The entire function can be written much shorter.

OlegPliss (changes requested):

runtime/driver.c
tylanphear: Yes, agreed.

runtime/driver.c
tylanphear: I think the enum made sense at one point during development, in order to make the code clearer (when we're growing vs shrinking), however I agree at this point there's no need for it (just commenting the two branches should suffice)

CuppoJava:
> Yes, I can easily split the PR into two (since the memory map stuff is mostly self-contained), however there are still some things (cleaning up headers) that are common to both.

Sounds good. I pre-emptively cherry-picked some headers you've pulled out and pushed them to master already. 

> * The interface of `launch_process` accepts a `stz_byte*` (stanza str) for the executable and a `stz_byte**` (array of stanza strs) for the arguments. I see that `escape-shell-command` turns a list of arguments into a single command line -- would the idea be to pass `NULL` for the arguments and just pass the entire command line for the executable?

Yes, feel free to change the C interface for `launch_process`. We can pass in the single command line directly (since that is what the windows API is expecting). The idea is to keep the C compatibility layer minimal. 

> * I am fine with this, however one note: `norm-path` seems to only handle the slash conversion, whereas `normalize_path` uses `GetFullPathName` to not only do that, but also to expand relative paths. I'm assuming we're fine with that (not sure what the behavior is on Linux)

In this case, if we use the proper call to `CreateProcess`, I believe we only want to keep the slash conversion part, because `CreateProcess` will do its own magic in converting the executable path into the actual exe to execute.

> * `CreateProcess` is actually a macro that expands to `CreateProcessA` (or `CreateProcessW`) depending on the value of `UNICODE`, so we are actually using `CreateProcessA`. The PATH-searching behavior is only triggered when `CreateProcess` is called with a `NULL` executable (right now we are passing one directly). Then Windows will perform a lookup based on the first argument of the command line.

Okay, let's use the built-in Windows path-searching algorithm. So it sounds like we should pass in `NULL` for the executable.

> I think a simple way to address multiple of your requests is to adjust the interface of `launch_process` on Windows to only take a command line (and to eliminate other unused arguments). Then we can just pass that to `CreateProcess` and have the Windows API do the heavy lifting for us.

Yup that sounds good to me. 

> * Yes, there is one platform-specific detail that complicates things. On Linux, when you request a memory mapping, the operating system reserves pages and commits them to memory automatically as-needed. On Windows, however, this commitment process must be done manually. Hence when we allocate our resizable segment, we `MEM_RESERVE` the entire maximum size, then `MEM_COMMIT` the minimum size (the initial size). As we resize, we have to take care to `MEM_COMMIT` and `MEM_DECOMMIT` as we grow/shrink respectively. This way the semantics of the POSIX implementation are preserved.

I see. Thanks for explaining. Do your best to shorten what you can then.

tylanphear:
OK, in light of the above, I've split these changes into the following 3 PRs: #96, #97, and #98. Closing.

PR #94, opened 2021-09-01
globally compute uses, defs, and elim-vars ahead of eliminating instructions

jackbackrack:
Eliminate-dead-code could miss dead code on functions with nested functions.  The solution is to compute the uses, defs, and vars to eliminate before eliminating the instructions.


CuppoJava (approved):
Looks good. THanks!

PR #96, opened 2021-09-02
Minor windows fixes

tylanphear:
Miscellaneous fixes and additions to make the Windows runtime compile successfully.

* Fix includes on Windows (should be lower-cased for cross-compilation on Linux)
* Compile out `fmalloc` completely when `FMALLOC` is not defined (relies on functionality that is only available on Linux)
* Fix overflowing integer literals (`long` is 32 bit on Windows so values over 32 bits overflow)

### DIFF
--- a/runtime/driver.c
+++ b/runtime/driver.c
@@ -25,7 +25,9 @@
 //       ====================
 
 //FMalloc Debugging
+#ifdef FMALLOC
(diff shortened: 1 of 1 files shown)


CuppoJava:
Looks good Tyler. Thank you!

PR #97, opened 2021-09-03
Windows support for resizable memory regions

tylanphear:
Implement resizable memory regions on Windows using the `Virtual*` family of functions.

Special care must be taken on Windows (as opposed to POSIX) to ensure that memory is explicitly committed before it is used, and decomitted when it is no longer in use. This is separate from the reservation of the memory region itself (which happens when the memory region is initialized.)

NOTE: This PR depends on #96 in order to compile


CuppoJava:
Looks good. Thank you Tyler!

PR #98, opened 2021-09-03
Windows process library support

tylanphear:
This implements the process library on Windows using the `CreateProcess` function and friends.

This implementation diverges from the original POSIX implementation in that there is no dedicated launcher process. This is due to a lack of the use of (and support for) `fork()`. However, almost all of the semantics of the original are honored.

One caveat is that the first element of the argument list passed to the `Process()` constructor is ignored in favor of the `filename` argument denoting the executable. This was done to mirror the POSIX API where `argv[0]` is merely conventional. It was also done out of necessity because the Windows API requires a command-line to be passed (executable + arguments concatenated) rather than an executable and a list of arguments (like POSIX)

NOTE: This PR relies on #96 in order to compile.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -63,10 +63,13 @@ protected extern stz_memory_unmap: (ptr<?>, long) -> int
 protected extern stz_memory_resize: (ptr<?>, long, long) -> int
 
 ;Process libraries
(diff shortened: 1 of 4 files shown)


CuppoJava:
Looks awesome. Great job Tyler!!!!

PR #101, opened 2021-09-03
Platform-specific error handling

tylanphear:
Introduce `platform-error-msg()`. This should be called in order to get the correct error message for the current platform (rather than calling `windows-error-msg()` or `linux-error-msg()` directly)


CuppoJava:
Looks good Tyler! Thank you!

PR #102, opened 2021-09-04
Implement Windows line prompter for REPL

tylanphear:
The regular line prompter uses `linenoise`, which is not available on Windows. In order to get basic functionality, this patch implements a simple line prompter that prints the prompt, takes input and then gives it back to the REPL. This doesn't get us any fancy behavior like command history, but it should get the job done for now.

Additionally, remove `linenoise.c` from the Windows finish script (shouldn't have been there in the first place)


CuppoJava (approved):
Thanks Tyler. The code looks good, I left some minor notes for educational purposes.

compiler/regular-line-prompter.stanza
```diff
@@ -0,0 +1,28 @@
+defpackage regular-line-prompter :
+  import core
+  import reader
+
+public defn LinePrompter (prompt: String, continue: String) -> Prompter:
+
+  val buffer = StringBuffer()
+  defn read-line (s: InputStream) -> String|False:
+    clear(buffer)
+    defn* loop () -> String|False :
```
CuppoJava, line 10: If you haven't seen the "named let" construct, here is a place that you could have written using it. It's not any shorter in this case, but sometimes it's very useful.

```
  clear(buffer)
  let loop () :
    val c = get-char(s)
    match(c:Char) :
      if c != '\n' :
        add(buffer,c)
        loop()
  to-string(buffer)
```

compiler/regular-line-prompter.stanza
```diff
@@ -0,0 +1,28 @@
...
+      match(get-char(s)):
+        (c:Char):
+          if c == '\n':
+            to-string(buffer)
+          else:
+            add(buffer, c)
+            loop()
+    loop()
+
+  var line-counter = to-seq(0 to false)
```
CuppoJava, line 20: The object (the seq) is mutable, but `line-counter` is not (i.e. it is always bound to the same seq.)

So you can declare this as:
```
val line-counter = to-seq(0 to false)
```

CuppoJava (approved):
Thanks Tyler. The code looks good, I left some minor notes for educational purposes.

PR #104, opened 2021-09-07
runtime: windows: don't call `CloseHandle()` twice on the same handle.

### LINKED ISSUE #103: Segfault on Windows when redirecting `STDOUT/STDERR` to the same stream
For example, if you compile the following test program:

```stanza
defpackage test:
  import core

defn main ():
  val args = command-line-arguments()
  Process(args[1], args[1 through false], STANDARD-IN, PROCESS-OUT, PROCESS-OUT)

main()
```

And run it, passing any valid executable as an argument (e.g. `./test gcc`), the test program seems to run fine in most cases. However, sometimes it will cause a segmentation violation. If run in the debugger (I used `gdb`), this segfault occurs reliably. This is due to an exception being thrown in the Windows runtime due to our closing a file handle twice.

Specifically, we redirect `STDOUT` and `STDERR` to the same handle (the `PROCESS-OUT` pipe), spawn the process with its `STDOUT` and `STDERR` set to the write end of the pipe, and then close those write ends in our process. However, since `STDOUT` and `STDERR` point to the same pipe, the latter action closes the `PROCESS-OUT` pipe twice, causing a segfault.

The same behavior can be observed if both streams are redirected to `PROCESS-ERR` (other combinations of redirections are not yet supported.)

tylanphear:
In order to keep good handle hygiene, we make sure to close the inherited handles that we pass into the process (`stdin_read`, `stdout_write` and `stderr_write`).

However, in the event that we redirect both output streams of a process to the same stream (i.e. `STDOUT/STDERR` to `PROCESS-OUT` or `STDOUT/STDERR` to `PROCESS-ERR`), we need to take care not to inadvertently close the same file handle twice. Since `stdout_write` and `stderr_write` point to the same handle, calling `CloseHandle()` on both causes us to close the same handle twice. This causes an exception in the Windows runtime, (sometimes) leading to a segfault upstream.

In order to prevent this scenario, we employ a trick similar to what we used to set up the file handles. We keep track of three pointers that correspond to a potential input/output/error handle, and then assign these to the corresponding redirected handle. We then close the pointers to valid handles, knowing that if we redirected both streams to the same handle, the second redirection would overwrite the handle value for the first, hence no double-close will occur.

Fixes #103

### DIFF
--- a/runtime/process-win32.c
+++ b/runtime/process-win32.c
@@ -188,6 +188,61 @@ static void setup_file_handles(
   }
 }
 
+static void close_and_null_if_not_null(PHANDLE handle_ptr) {
+  if (handle_ptr != NULL && *handle_ptr != NULL) {
+    CloseHandle(*handle_ptr);
+    *handle_ptr = NULL:
+  }
+}
+
+// Close all inherited file handles that we passed into the child process. This may seem
+// overcomplicated, but is necessary due to the fact that we allow multiple streams to be redirected
+// to the same stream. For example, we allow `STDOUT` and `STDERR` to be redirected to both be
+// redirected to `PROCESS-OUT` (or `PROCESS-ERR`) simulataneously. In such a case, we have to be
+// careful not to naively call `CloseHandle()` on both streams, since that would cause the pipe end
+// to be closed twice, causing an exception in the Windows API (and likely a segfault upstream.)
+static void close_inherited_file_handles(
+    stz_int input, stz_int output, stz_int error,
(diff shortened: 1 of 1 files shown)


OlegPliss (changes requested):
I don't think it is a correct fix. You are not supposed to pass the same handle twice as different arguments to the process. Handles have to be duplicated by the caller.

tylanphear:
I left a comment in response to your original comment (https://github.com/StanzaOrg/lbstanza/issues/103#issuecomment-914709564) (which I won't duplicate here, just linking for documentation). I will investigate a solution along those lines instead.

PR #105, opened 2021-09-08
Windows runtime: Duplicate pipe handles we pass into the child process

### LINKED ISSUE #103: Segfault on Windows when redirecting `STDOUT/STDERR` to the same stream

tylanphear:
When we create pipes for redirecting a child process's input/output/error streams, we have to take special care to duplicate
any pipe ends that may be shared by other streams. This is to prevent two things:

1. Two streams receiving the same handle to a pipe end (technically invalid)
2. Closing the same handle to a pipe end twice when attempting to close the ends of the stream that we passed into the child (causes an exception in the Windows runtime)

For example, with the original code, if we were to redirect both `STDOUT` and `STDERR` to `PROCESS-OUT`, we would end up with both `STDOUT` and `STDERR` in the child process having the same handle to their end of the `PROCESS-OUT` pipe. We would then call `CloseHandle()` on those handles in the parent process, only to inadvertently close the same handle twice. Duplicating the handles ensures that both streams have a handle to the same pipe end, but not the same handle, resolving both of the above issues.

Fixes #103

### DIFF
--- a/runtime/process-win32.c
+++ b/runtime/process-win32.c
@@ -97,12 +97,12 @@ static void create_pipe(PHANDLE read, PHANDLE write, PipeType type) {
   SetHandleInformation(*our_end, HANDLE_FLAG_INHERIT, 0);
 }
 
-static HANDLE duplicate_standard_handle(int handle) {
+static HANDLE duplicate_handle(HANDLE handle) {
   HANDLE ret;
 
   if (!DuplicateHandle(
         /* hSourceProcessHandle */ GetCurrentProcess(),
-        /* hSourceHandle        */ GetStdHandle(handle),
+        /* hSourceHandle        */ handle,
         /* hTargetProcessHandle */ GetCurrentProcess(),
         /* lpTargetHandle       */ &ret,
         /* dwDesiredAccess      */ 0,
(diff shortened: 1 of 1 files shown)


CuppoJava:
Looks much better. Thank you!

PR #108, opened 2021-09-12
Added comments during reading.

CuppoJava:


core/core.stanza
```diff
@@ -1332,81 +1337,131 @@ lostanza defn iterate-marked (start:ptr<?>, limit:ptr<?>,
...
-    if marking-stack-not-empty(vms) == 0 : fatal("Marking stack underflow.")
+    if marking-stack-empty(vms) : fatal("Marking stack underflow.")
   val value = [vms.new-heap.stack-top]
   vms.new-heap.stack-top = vms.new-heap.stack-top + BYTES-IN-LONG
   return value as ptr<long>
 
+;Returns 1L if the marking stack is not empty.
+;The marking stack grows downwards, so it is not empty
+;if the top is less than the bottom. 
 lostanza defn marking-stack-not-empty (vms:ptr<VMState>) -> long :
```
CuppoJava, line 1379: I added some more functions for testing the state of the marking stack.

core/core.stanza
```diff
@@ -1332,81 +1337,131 @@ lostanza defn iterate-marked (start:ptr<?>, limit:ptr<?>,
...
   ;No meaningful return value
   return false
 
-lostanza defn push (p:ptr<long>, vms:ptr<VMState>) -> ref<False> :
+;Push a new pointer onto the marking stack.
+;Note that the marking stack grows downwards.
+lostanza defn push-to-marking-stack (p:ptr<long>, vms:ptr<VMState>) -> ref<False> :
   #if-not-defined(OPTIMIZE) :
-    if vms.new-heap.stack-top > vms.new-heap.stack-start : fatal("Marking stack overflow.")
+    if marking-stack-full(vms) : fatal("Marking stack overflow.")
```
CuppoJava, line 1362: I changed this behaviour.

I believe the original condition is not correct. Can you confirm?

OlegPliss: Yes, you are right!

core/core.stanza
```diff
@@ -1332,81 +1337,131 @@ lostanza defn iterate-marked (start:ptr<?>, limit:ptr<?>,
...
   return false
 
+;Given the pointer p to an already-marked object,
+;mark all the objects that it references and traverse the object graph.
+;Marks as many objects as possible given the size of the marking stack.
 lostanza defn continue-marking (p:ptr<?>, vms:ptr<VMState>) -> ref<False> :
   #if-not-defined(OPTIMIZE) :
     if test-mark(p, vms) == 0 : fatal("Marked object expected.")
-  push(p, vms)
+    if marking-stack-not-empty(vms) : fatal("Marking stack expected to be empty.")
```
OlegPliss: This condition is not essential. Current implementation would successfully handle some preexisting objects in the marking stack. So I would not make it a pre-condition.

CuppoJava: Okay. sounds good. I'll remove this one.

PR #109, opened 2021-09-13
jb/typed resolve methods

jackbackrack:
Switch to using type inferred variables for the method resolution.  In so doing, need to update type inference to allow for it to be called later.  Specifically, added support for boxes.  Also, because beta reduction is removing basic block-specific variables, added capability for typing variables differently in different basic blocks.  From there changed resolve-methods to use inferred types for method resolution.  Also added the ability to resolve match instructions based on inferred types.  Finally, needed to clean up live instructions that were pointing to inlined functions.


CuppoJava:
I just released Stanza `0.14.18` which contains the new inference engine, and pulled in and refactored the changes you made to `resolve-methods` to use the new inferred types. Can you verify whether the new inference engine is accurate enough for the passes you're running?

I still need to merge the simple-inline `ELive` fix. Thank you.

I am confused about the match resolution. There is an existing pass called `resolve-matches`, what is the relation between the new match resolution code and that pass?

jackbackrack:
resolve-matches is different.  it merely makes sure that the match arguments are fully covered by the match branches so that the match failure guard can be removed.  on the other hand, my match resolution actually partially evaluates match instructions to prune away unnecessary branches and if only one remains to turn the match into a goto.

CuppoJava:
I see now. Thanks for explaining! 

For computing the pruning, can you change the algorithm to use the `all-solns` function in the dispatch dag? The key change to get it to behave like a `match` instead of a `multi` is to pass `false` to the `topological?` argument in `compute-dispatch-dag`. 

Your strategy can then be extended  for `EDispatch` statements as well following this strategy.

PR #110, opened 2021-09-13
jb/remove boxes

jackbackrack:
Optimization pass to remove boxes if all box usages are in the same body.


compiler/stz-el.stanza
```diff
@@ -2501,6 +2506,78 @@ defn union (input-ts:Seqable<EType>) :
...
+;============================================================
+;===================== REMOVE-BOXES =========================
+;============================================================
+;Currently boxes are used to hold variables that are written to
+;In this pass, we remove boxes for those boxes that are only
+;accessed locally in a single body, since the ELIR supports
+;writing to local vars.  This permits a number of follow on
+;optimizations to trigger as well.  
+defn remove-boxes (epackage:EPackage, gvt:VarTable) -> EPackage :
+  val non-box-reads = IntTable<Int>(0)
```
CuppoJava: Do we need the integer count here for `non-box-reads`? Or is an `IntSet` enough?

compiler/stz-el.stanza
```diff
@@ -2501,6 +2506,78 @@ defn union (input-ts:Seqable<EType>) :
...
+  val local-boxes = IntSet()
+  val box-types = IntTable<EType>()
```
CuppoJava: The logic for the `box-types` table is tough and annoying because it is information that we used to have that was thrown out. 

Instead of this, let's preserve the content type of the box when it was first created, and store this in the `EBox` instruction.

compiler/stz-el.stanza
```diff
@@ -2501,6 +2506,78 @@ defn union (input-ts:Seqable<EType>) :
...
+
+  ;Count all reads that are not box reads that is
+  ;that are not through EBox/EBoxGet/EBoxSet
+  defn count-non-box-reads (e:ELItem) :
+    let loop (e:ELItem = e) :
+      match(e) :
+        (e:EVar|EVarLoc) :         update(non-box-reads, {_ + 1}, n(e))
+        (e:EBox|EBoxGet|EBoxSet) : false
```
CuppoJava: The `z` field in `EBoxSet` is still a potential reference to a box.

compiler/stz-el.stanza
```diff
@@ -2501,6 +2506,78 @@ defn union (input-ts:Seqable<EType>) :
...
+  ;Also collect the box types so as to update the associate local var type
+  defn collect-local-boxes (e:ELItem, vt:VarTable) :
+    let loop (e:ELItem = e) :
+      match(e) :
+        (e:EBox) :    add(local-boxes, n(x(e))) when non-box-reads[n(x(e))] == 0
+        (e:EBoxGet) : box-types[n(y(e) as EVar)] = type(vt, n(x(e)))
+        (e) :         do(loop, e)
+
+  ;Translate all local boxes and their accesses to regular vars and regular vars and defs
+  defn walk-body (e:EBody, vt:VarTable) -> EBody :
```
CuppoJava: Should the global tables `local-types`, `box-types`, and `non-box-reads` be reset for analyzing each body?

compiler/stz-el.stanza
```diff
@@ -2501,6 +2506,78 @@ defn union (input-ts:Seqable<EType>) :
...
+    defn local-box? (n:Int) : local-boxes[n] and key?(box-types, n)
+
+    defn walk (e:ELItem) -> ELItem :
+      match(e) :
+        (e:EBox) :
+          EDef(x(e), y(e)) when local-box?(n(x(e))) else e
+        (e:EBoxGet) :
+          EDef(x(e), y(e)) when local-box?(n(y(e) as EVar)) else e
+        (e:EBoxSet) :
+          EDef(EVarLoc(n(y(e) as EVar)), z(e)) when local-box?(n(y(e) as EVar)) else e
```
CuppoJava: I believe this should be an `EStore` instruction, not an `EDef`.

compiler/stz-el.stanza
```diff
@@ -2501,6 +2506,78 @@ defn union (input-ts:Seqable<EType>) :
...
+      ELocal(n(local), box-types[n(local)], true) when local-box?(n(local)) else local
+
+    sub-locals(sub-ins(e, ins*), locals*)
+
+  defn fold-texp (x:ETExp, vt:VarTable) :
+    val result = let loop (x:ELBigItem = x) :
+      match(x) :
+        (e:EBody) :
+          ;Only do local box translation if no nested functions of objects
+          if empty?(localfns(e)) and empty?(localobjs(e)) :
```
CuppoJava: It's worth getting this pass working on general functions as well. A common idiom is to do something like this:

```
defn mymain () :
  var x = 0
  defn increment-x () : x = x + 1
  defn big-nested-function () :
    some-function(1)
    ...
    some-function(2)
  increment-x()
  ... my code ...
  increment-x()
```

where we use some small helpers to manipulate local state. During inlining, all of those helpers will get inlined, so the variable will be free to be unboxed.

CuppoJava (approved):
Thanks for the comments. Had some questions about the tables and the operations.

compiler/stz-el.stanza
jackbackrack: Are you saying that there are functions that don't have EBody's?  I'm not sure I understand what a general function is.

compiler/stz-el.stanza
jackbackrack: Okay.

compiler/stz-el.stanza
jackbackrack: Do I need to?  It takes time to clear them I think.  I thought all IDs were unique.

compiler/stz-el.stanza
jackbackrack: OK.  I agree. I'll look into that.

compiler/stz-el.stanza
jackbackrack: You're right.  IntSet is enough.

compiler/stz-el.stanza
jackbackrack: Why won't the current pass work on this (once everything is inlined)?

CuppoJava: In the example above, `increment-x` is a very small function that is likely to be inlined. So after inlining, `increment-x` won't exist anymore, but `big-nested-function` will continue to exist. 

However, because `increment-x` has been inlined, we will know that the box created for the variable `x` won't escape the local scope and thus can be transformed using your box-removal logic. So it is worthwhile to make this happen.

The current pass won't do this because `big-nested-function` is still there, and hence doesn't pass the `Only do local box translation if no nested functions of objects` condition.

compiler/stz-el.stanza
CuppoJava: That's true, they are unique. But if you don't clear them, your table will eventually be filled with the every identifier in the program. That will be quite large.

compiler/stz-el.stanza
jackbackrack: I understand better now.  The question now is how do I track if all box refs come from same ebody?   I could imagine tagging each ref with the id of the body but ebody's don't have ids.  The other way would be to walk nested functions first and then when walking body instructions see whether box has been refd before.  It all seems doable but tricky and somewhat inefficient (with multiple passes etc) was wondering if you have a particular suggestion for how you thought the analysis might go?

compiler/stz-el.stanza
jackbackrack: So that brings up the question of whether clearing an existing table is better than creating a new one?

CuppoJava: I think the performance difference between the two is negligible.

compiler/stz-el.stanza
CuppoJava: You can do it in a single pass if you analyze the nested functions first, and pass out the information about non-locals you reference in each function as you go.

Alternatively, you can also call the analyze-freevars function and just accept one more pass. This is way I would go. 

If the passes are linear, I would like to eat the cost of the passes for now for the sake of code clarity.

PR #111, opened 2021-09-14
peephole prune redundant live vars, and eliminate defs to unused literals

jackbackrack:
Improve dead code elimination so that ELIR is more efficient and readable.


compiler/stz-el.stanza
```diff
@@ -2732,39 +2735,56 @@ defn eliminate-dead-code (epackage:EPackage) -> EPackage :
...
+              key?(use-table, n(v))
+        else : default-check()
 
     ;A variable can be eliminated if all instructions that define
     ;it can be eliminated.
     to-intset $ for entry in def-table seq? :
       if all?(eliminate?, value(entry)) : One(key(entry))
       else : None()      
 
+  ;Pending live vars to be emitted next.
```
CuppoJava, line 2292: Hi Jonathan, can you explain what this code is doing, and whether it is appropriate to be considered as part of dead-code elimination, or whether it should be moved to a different pass?

compiler/stz-el.stanza
```diff
@@ -105,6 +105,9 @@ defn lower (epackage:EPackage, optimize?:True|False) -> EPackage :
   run-pass("Lift Closures", lift-closures, "closurelifted", false)
   run-pass("Lift Type Objects", lift-type-objects, "typelifted", false)
 
+  if optimize? :
```
CuppoJava: Remove debugging code.

compiler/stz-el.stanza
```diff
@@ -2732,39 +2735,56 @@ defn eliminate-dead-code (epackage:EPackage) -> EPackage :
   defn vars-to-eliminate (def-table:IntListTable<EIns>, use-table:IntTable<UsageType>) -> IntSet :
     ;Returns true if the given instruction can be eliminated.
     ;Note that for 'EDef x = v' instructions, we consider
-    ;ELive as a proper usage so that object live ranges are preserved.
+    ;ELive as a proper usage so that object live ranges are preserved,
```
CuppoJava: Change the condition: We consider ELive as a proper usage only for EDef x = v instructions, where v is an EVar.

CuppoJava (changes requested):
Request for clarification.

compiler/stz-el.stanza
jackbackrack: Currently, you're checking to see if EDef varlocs are used either as args or in ELive instructions.  This change now says that EDefs with literal y's are removed if varlocs are only used in live instructions (as opposed to args).  Without this you end of with a bunch of useless literal defs.  We talked about this and you agreed.  I changed the function name to make it clearer.

compiler/stz-el.stanza
jackbackrack: done.

compiler/stz-el.stanza
CuppoJava: Sorry, I meant the peephole stuff. What is that for?

jackbackrack: The peephole stuff is for removing extra live statements and consolidating them into a single live statement.  For instance, I would see many successive ELive statements with redundant variables in them.  This collapses them into a single ELive statement with the union of those variables.  It just cleans up the ELive statements so they don't dominate the EL-IR.

CuppoJava: Okay sounds good. Can you look into refactoring/cleaning up the logic for this pass so the logic is easier to understand?

jackbackrack: i worked pretty hard to make the live var pruner code simple.  i'm not sure how much refactoring is possible in 12 lines of code.  perhaps i can add comments to make it simpler.  i could split it out into a separate sweep through the instructions but then it would be slower...

jackbackrack: ok, i've pushed a new version that splits out consolidate live vars.

PR #115, opened 2021-09-15
Jb/fold box unbox match if

jackbackrack:
Fold away matches and ifs that are essentially box/unbox noops.


CuppoJava (commented):
Can you remove `check-cmp-arg?` and give me the reproduction instructions for crashing the backend? I can fix that up and clean up this pass.

PR #116, opened 2021-09-20
Implement LivenessTracker

OlegPliss:
LivenessTrackers have to be reimplemented, otherwse all LivenessTrackers are reachable from LIVENESS-HANDLERS, and so are the references.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -1207,8 +1207,7 @@ lostanza defn set-mark (p:ptr<?>, vms:ptr<VMState>) -> ref<False> :
   return false
 
 ;Marks the given heap pointer in the state's heap bitset.
-;Returns a non-zero value if the pointer was previously marked.
-;Returns zero if the pointer was not previously marked.
+;Returns 0L when the pointer was previously marked, otherwise 1L.
 lostanza defn test-and-set-mark (p:ptr<?>, vms:ptr<VMState>) -> long :
   val index = bit-index(p, vms)
   val address = bit-address(index, vms)
@@ -1216,7 +1215,7 @@ lostanza defn test-and-set-mark (p:ptr<?>, vms:ptr<VMState>) -> long :
 
   val old-value = [address]
   [address] = old-value | mask
-  return old-value & mask
+  return (old-value & mask) == 0
 
 ;Clears the mark for the given heap pointer in the state's heap's bitset.
 lostanza defn clear-mark (p:ptr<?>, vms:ptr<VMState>) -> ref<False> :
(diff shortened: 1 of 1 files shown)


core/core.stanza
```diff
@@ -1558,6 +1557,46 @@ lostanza defn mark-reachable-objects (vms:ptr<VMState>) -> ref<False> :
...
+  LIVENESS-TRACKERS = tracker
+  return tracker
+
+public lostanza defn value (t:ref<NewLivenessTracker>) -> ref<False|Unique> :
+  return t.value as ref<False|Unique>
+
+;Global list of live LivenessTrackers
+lostanza var LIVENESS-TRACKERS:ref<False|NewLivenessTracker> = false
+
+lostanza defn scan-liveness-trackers (vms:ptr<VMState>) -> ref<False> :
```
CuppoJava, line 1585: From my understanding, this algorithm changes the current semantics of `LivenessTracker` in a very subtle way. Can you confirm whether I understood the algorithm correctly?

The current GC treats LivenessTrackers themselves as an ordinary object. They are supposed to be garbage collected if the LivenessTracker itself is no longer reachable. So in the current implementation, the following will run forever without running out of memory.

```
defstruct MyUniqueObject :
  value: String
  
defn main () :
  val obj = MyUniqueObject("Hello world")
  while true :
    LivenessTracker(obj)
  
main() 
```

In the new algorithm, a LivenessTracker is garbage-collected only if the value it is wrapping is no longer live. Thus the above code will eventually die with an Out-Of-Memory error, since `obj` is always live.

CuppoJava (commented):
Question about algorithm.

core/core.stanza
OlegPliss: Yes, this is correct. But it can be easily fixed. `tail: ref<?>` and `LIVENESS-TRACKERS:ref<False|NewLivenessTracker>`can be replaced with `ptr<NewLivenessTracker>`. Unmarked elements of the list can be disposed in `scan-liveness-trackers`.

CuppoJava: Thanks for replying. That's a very straightforward fix! Thank you!

OlegPliss: Fixed.

CuppoJava:
Thanks Oleg!

This PR has been merged with my own comments, as usual, plus the following fixes and modifications:

1) FIX: Change cast to `addr!` operator.
```
LIVENESS-TRACKERS = tracker as ptr<NewLivenessTracker>
```

has been changed to:
```
LIVENESS-TRACKERS = addr!([tracker])
```

The first one takes `tracker` which is a `ref<NewLivenessTracker>` and interprets its bit representation (with the additional tag bits, etc.) as a pointer.

The second one computes the memory address that the stored `NewLivenessTracker` struct is stored at. 

2) FIX: Subtract header word to retrieve heap pointer before calling `test-mark`. 

```
val tracker = [p]
if test-mark(tracker, vms) == 0 :
  ...
```

has been changed to:

```
val tracker = [p]

;Compute the pointer to the start of the object by decrementing
;the size of the header word. This pointer is what is marked
;in the bitset. 
val tracker-obj = (tracker as ptr<?>) - sizeof(long)

if test-mark(tracker-obj, vms) == 0 :
  ...
```

3) MODIFICATION: Remove usage of `addr!` operator.

```
[addr!(tracker.value)] = false-marker()
```

has been changed to:

```
tracker.value = false-marker()
```

PR #120, opened 2021-09-22
Solid prefix detection

OlegPliss:
Solid prefix is a maximum sequence of live objects at the bottom of collection area (i.e. the entire heap is absence of generations). It can be of significant size. Objects of solid prefix do not move, so there is no need to build and lookup a break table for them.


CuppoJava:
Looks great. Merged!

PR #122, opened 2021-09-24
Break table construction

OlegPliss:
Break table is necessary for relocation of pointers and compaction phases.

"Break" is a conventional name for a maximum sequence of unmarked objects starting at given address. A break can consist of one or more longs. This garbage collector does not mandate 2 longs (or 16 bytes) minimum object size. One long is sufficient. 

A break contains a pointer to the first marked object above it and a pointer to the next break. The pointers are aligned. To fit into single long, special encoding is used for the first marked object above. In a single-long break the address of the first marked object above is designated by 1 in the lowest bit of the pointer to the next break.

Break table is represented by an implicit list of breaks and a hashed root to accelerate lookup.

The list of breaks starts at the lowest break. `compaction-start` conveniently points to this location. Compaction phase moves a segment of reachable object downwards by the sum of break lengths below the segment.

To avoid computing this sum for given object from the list entry, compaction area is divided into equally sized buckets. Break table root consists of entries each of which contains a pointer to the start of the lowest break is the bucket and accumulated sum of breaks below it.  To compute a sum of breaks below given object, we locate its bucket, prime the counter with the stored accumulated breaks below the bucket and continue to accumulate breaks below the object starting with the stored break.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -1633,7 +1633,7 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
 ;Scans the heap from given start to heap top until first marked object found.
 ;Returns address of the first unmarked object or the heap top if it is not found.
 lostanza defn skip-unmarked (start:ptr<long>, vms:ptr<VMState>) -> ptr<long> :
-  ;Precondition: Ensure that given start is below heap top
+  ;Precondition: Ensure that given start is below or equal to heap top.
   val heap-top = vms.new-heap.top
   #if-not-defined(OPTIMIZE) :
     if start > heap-top : fatal("Invalid range.")
@@ -1769,7 +1769,7 @@ lostanza defn create-break-table (compaction-start:ptr<long>, vms:ptr<VMState>) 
   ;- index: the index of the last bucket whose entry has been filled.
   ;- limit: the top of the heap
   ;- offset: the accumulated offset that all objects in the current bucket
-  ;  should be shifted by. Guaranteed to be non-positive because
+  ;  should be shifted by. Guaranteed to be strictly negative because
   ;  objects can only move downwards during compaction phase.
   ;- current-unmarked: points to the next break in the heap.
   var index:long = -1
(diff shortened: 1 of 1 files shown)


core/core.stanza
```diff
@@ -1623,15 +1626,100 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
...
+lostanza defn read-live-range (location:ptr<long>) -> LiveRange :
+  val location0 = location[0]
+  if location0 & 1L : return LiveRange{addr(location[1]), (location0 - 1) as ptr<long>}
+  else : return LiveRange{location0 as ptr<long>, location[1] as ptr<long>}
+
+lostanza deftype BreakTableEntry :
+  var first-break-location:ptr<long>
+  var accumulated-offset:long
+
+lostanza defn log2-bucket-size (break-table-length:long, vms:ptr<VMState>) -> int :
```
CuppoJava, line 1724: I had difficulty understanding what this function was computing, and added the following comments. Can you confirm whether it is accurate? 
```
  ;Goal: Compute minimum n such that:
  ;      2^n * break-table-length >= total-compaction-area
  ;Equivalent to: Compute minimum n such that:
  ;               total-compaction-area / 2^n  <=  break-table-length
  var count:int = 0
  var compaction-area-size:long = vms.new-heap.top - vms.new-heap.compaction-start - 1L
  while compaction-area-size >= break-table-length :
    count = count + 1
    compaction-area-size = compaction-area-size >> 1
  return count
```

Is the following algorithm equivalent? I find it easier to understand:
```
  ;Goal: Compute minimum n such that:
  ;      2^n * break-table-length >= total-compaction-area
  ;Equivalent to: Compute minimum n such that:
  ;               total-compaction-area / 2^n  <=  break-table-length
  var count:int = 0
  var compaction-area-size:long = vms.new-heap.top - vms.new-heap.compaction-start
  while compaction-area-size > break-table-length :
    count = count + 1
    compaction-area-size = compaction-area-size >> 1
  return count
```

OlegPliss: It computes `ceil(log2((vms.new-heap.top - vms.new-heap.compaction-start) / break-table-length)`.

 These two implementations seem equivalent. The last one is one integer decrement shorter, so I would borrow it :)

OlegPliss: Apparently, the functions are not equivalent. I.e. for `vms.new-heap.top - vms.new-heap.compaction-start` == 15, `break-table-length` == 1 they return 4 and 3 respectively, and 4 is correct result.

CuppoJava: Thank you for replying Oleg! I'll leave the code alone then, though I admit I still don't quite understand how it works.

PR #128, opened 2021-09-28
Break table reading

CuppoJava:
Added comments during code reading of break table implementation.


core/core.stanza
```diff
@@ -1627,85 +1631,194 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
   return heap-top
 
 ;Scans the heap from given start to heap top until first marked object found.
-;Returns address of the first unmarked object or the limit if it is not found.
+;Returns address of the first unmarked object or the heap top if it is not found.
 lostanza defn skip-unmarked (start:ptr<long>, vms:ptr<VMState>) -> ptr<long> :
-  val limit = vms.new-heap.top
+  ;Precondition: Ensure that given start is below heap top
```
OlegPliss: below or equal heap top?

core/core.stanza
```diff
@@ -1627,85 +1631,194 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
...
+    return LiveRange{live, dead}
+  else :
+    ;Case break >= 16 bytes.
+    val live = location0 as ptr<long>
+    val dead = location[1] as ptr<long>
+    return LiveRange{live, dead}
+
+;The break table considers the heap broken up into equal-sized buckets.
+;Each bucket stores its information in a BreakTableEntry.
+;- first-break-location: the address right after the first live range in this break.
```
OlegPliss, line 1714: This is not correct. First break is a location of the lowest live range which [live, dead) semi-open interval intersects with the bucket address range.

core/core.stanza
```diff
@@ -1627,85 +1631,194 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
...
-  ;so the object address offset can only be negative.
+  ;Construct break table using the memory of the marking stack.
+  val break-table = vms.new-heap.stack-start as ptr<BreakTableEntry>  
+
+  ;Fill the break table and annotate the breaks with the live ranges.
+  ;- index: the index of the last bucket whose entry has been filled.
+  ;- limit: the top of the heap
+  ;- offset: the accumulated offset that all objects in the current bucket
+  ;  should be shifted by. Guaranteed to be non-positive because
+  ;  objects can only move downwards during compaction phase.
```
OlegPliss, line 1773: Actually it is guaranteed to be strictly negative. Break table is built for compaction area only. Solid prefix is excluded from it.

core/core.stanza
```diff
@@ -1627,85 +1631,194 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
...
     val start-index = break-table-index(next-marked, vms)
-    if index < start-index : index = start-index
     val end-index = break-table-index(next-unmarked - 1, vms)
+
+    ;Advance the index to be at least the starting index of the
+    ;live range.
+    if index < start-index : index = start-index
+
+    ;For all buckets that the live range spans, fill in the same
+    ;BreakTableEntry.
```
OlegPliss: ... or none if index exceeds end index.

core/core.stanza
```diff
@@ -1627,85 +1631,194 @@ lostanza defn skip-and-clear-marked (start:ptr<long>, vms:ptr<VMState>) -> ptr<l
...
+;The minimum size of a break is 8 bytes long, but the live range struct is
+;in general 16 bytes long.
+;
+;Encoding:
+;Case break >= 16 bytes: then just directly store the live range.
+;Case break == 8 bytes: only store the 'dead' field, tagged with 1.
+;
+;Since heap addresses are always 8-byte aligned, the tagged 1 will tell us
+;which encoding we are using. 
+lostanza defn write-live-range (location:ptr<long>, r:LiveRange) -> ref<False> :
```
OlegPliss, line 1676: I wonder if our current compiler is good enough to pass LiveRange in registers? Performance of this function is critical.

core/core.stanza
CuppoJava: Fixed.

core/core.stanza
CuppoJava: Can you explain this some more? I had trouble understanding this part. Does `first-break-location` point to a live object or dead object?

Later in the code, I saw this:
```
      break-table[index] = BreakTableEntry{next-unmarked, offset}
```

Which seems like it points to a dead object.

core/core.stanza
CuppoJava: Thank you. Fixed.

core/core.stanza
CuppoJava: Thank you. Fixed.

core/core.stanza
CuppoJava: Yes. The two different usages compile down to the same code.

core/core.stanza
OlegPliss: It always points to a `LiveRange` record which is stored in the beginning of a sequence of dead objects.

CuppoJava: Ah okay, I see. 

There's one behaviour that I am confused by: What is the layout of the very very first break?

Suppose the following addresses:
- the heap starts at 0x1000000, and the object at that location is dead. 
- the next live range extends from [0x1001000, 0x1003000), and this live range is entirely in the first bucket.

From the code, it seems that:
1) There is a LiveRange struct written directly at 0x1000000. 
2) There is another LiveRange struct written at 0x1003000.
3) The first BreakTableEntry, `break-table[0]`, has its `first-break-location` set to 0x1003000. 

Is that what you expect? 

Point 3 is the one that stands out to me, that I would like to confirm.

OlegPliss: That's correct. 

`break-table[0].first-break-location` == 0x1003000
`break-table[0].accumulated-offset` == -0x1000

In this bucket an address below `0x1003000` is relocated by `-0x1000` offset.

PR #136, opened 2021-11-03
Remove global buffer

PhilippeFerreiraDeSousa:
I got several errors on elements accumulating in `SORT-BUFFER` from different `qsort` calls, the last of which I have no idea why. So here is the fix.
The decision to have a global buffer is surprising in the first place.

### DIFF
--- a/core/core.stanza
+++ b/core/core.stanza
@@ -5091,31 +5091,25 @@ public defn qsort!<?T,?S> (key:T -> ?S&Comparable<S>, xs:IndexedCollection<?T>) 
 ;                        Non-Destructive Sorting
 ;                        =======================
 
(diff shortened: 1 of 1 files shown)


OlegPliss:
The change looks good to me. What is not clear is how the elements can accumulate in SORT-BUFFER. There must be a hidden bug somewhere. Do you have a small reproducer for the problem?

PhilippeFerreiraDeSousa:
Yes call `qsort(xs, is-less?)` where `is-less?` itself calls `qsort`. SO you do a sort while another sort is already ongoing.

bpellenard:
@PhilippeFerreiraDeSousa, could you please add a unit test that would reproduce this error?

CuppoJava:
Merged.

PR #140, opened 2021-11-16
Handle directory creation/deletion on Windows appropriately

m-hilgendorf:
This PR introduces some changes to fix directory creation/deletion on windows. 

1. The C binding for `mkdir` is a function called `int _mkdir (const char*)` under mingw, which has one fewer argument. (mkdir exists, with the same function signature, but is marked deprecated)
2. `int remove (const char*)` fails with `Permission denied` when called on Windows with the argument pointing to a directory. 
3. A binding to `rmdir` is introduced, and `delete-file` is modified on Windows to check for the file type (if it exists) and call `rmdir` if the file type is `DirectoryType`.


CuppoJava (approved):
Thanks very much!

PR #142, opened 2021-11-19
Init Github Nix Workflow

SCOTT-HAMILTON:
Here is a simple workflow for continuous building of master after each commits on a linux machine leveraging [Nix](https://nixos.org/manual/nix/stable/).

### DIFF
--- a/nix/lbstanza-bin/default.nix
+++ b/nix/lbstanza-bin/default.nix
@@ -6,11 +6,11 @@
 
 stdenv.mkDerivation rec {
   pname = "lbstanza-bin";
-  version = "0.14.33";
+  version = "0.15.1";
 
   src = fetchzip {
     url = "http://lbstanza.org/resources/stanza/lstanza_${lib.replaceStrings ["."] ["_"] version}.zip";
-    sha256 = "1az9pddxh5kjqkkxsl3ffc23zjawilpisnvy9sy7ymnsp3b791lf";
+    sha256 = "sha256-Ogfe5HX+fk3Tth5J07fWngyHnQjbkYBxKObqsNY8spc=";
     stripRoot=false;
   };
 
--- a/nix/lbstanza-bin/default.nix
+++ b/nix/lbstanza-bin/default.nix
@@ -1,0 +1,35 @@
+{ lib
+, stdenv
+, fetchzip
+, autoPatchelfHook
+}:
+
+stdenv.mkDerivation rec {
+  pname = "lbstanza-bin";
+  version = "0.14.33";
+
+  src = fetchzip {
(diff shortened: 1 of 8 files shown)


nix/lbstanza-bin/default.nix
```diff
@@ -0,0 +1,35 @@
+{ lib
+, stdenv
+, fetchzip
+, autoPatchelfHook
+}:
+
+stdenv.mkDerivation rec {
+  pname = "lbstanza-bin";
+  version = "0.14.33";
```
CuppoJava: Thank you for this Scott! I'm not familiar with NIX syntax. 

Is there a way to pull this version from `compiler/stz-params.stanza`? I'm don't want to accidentally forget and accidentally leave this workflow unsynchronized.

CuppoJava (commented):
Thank you!

CuppoJava:
Thanks Scott! Left one comment on your PR, and closing this issue.

nix/lbstanza-bin/default.nix
SCOTT-HAMILTON: This CI isn't meant to be used against latest stable release but rather against master, it's meant to test every commit.

The way it does it is by first using this binary stanza to build master. So this version number isn't what we are testing, it's the version of the stanza compiler we are using to try and build master.

Do we really need to always use the latest stable binary release indicated in `compiler/stz-params.stanza` in order to build master ?

CuppoJava:
Hmm. I meant to close the issue, not the PR. But I guess the two are linked.

Thanks for the explanation. It would be a little more robust if it used the binary release indicated in `compiler/stz-params.stanza` in master. There are two types of changes that cause this:

1) Sometimes we change the code generation. In this case, we need the code generated from the compiler to be consistent with what is expected by `core`. 

2) Sometimes (very rarely) we change some macro definition and then use the updated definition in the implementation of the compiler. In this case, we will need the updated compiler in order to parse its own sources.

SCOTT-HAMILTON:
> Hmm. I meant to close the issue, not the PR. But I guess the two are linked.
> 
> Thanks for the explanation. It would be a little more robust if it used the binary release indicated in `compiler/stz-params.stanza` in master. There are two types of changes that cause this:
> 
> 1. Sometimes we change the code generation. In this case, we need the code generated from the compiler to be consistent with what is expected by `core`.
> 2. Sometimes (very rarely) we change some macro definition and then use the updated definition in the implementation of the compiler. In this case, we will need the updated compiler in order to parse its own sources.

Ok, I've pushed one commit that makes the CI automatically update `nix/lbstanza-bin` to the version indicated in `compilers/stz-params.stanza` if needed. This will generate automatic commits every time the version in `compilers/stz-params.stanza` changes. The workflow is configured to commit under my name, works on my fork but not sure if it will work once merged upstream.

PR #150, opened 2022-03-30
Refactored indexing algorithm and DefinitionsDatabase

m-hilgendorf:
- Added support for visibility modifiers in the definitions database
- Altered algorithm to flatten IL-IR into definitions and their visibility
- Split indexing functions into helpers to make it easier to test
- Added unit tests to check each step of the indexing process.

### DIFF
--- a/compiler/stz-defs-db-serializer.stanza
+++ b/compiler/stz-defs-db-serializer.stanza
@@ -3,6 +3,7 @@ defpackage stz/defs-db-serializer :
   import core
   import collections
   import stz/serializer
+  import stz/visibility
 
 ;===============================================================================
 ; =============================== Types ========================================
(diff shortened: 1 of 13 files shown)


compiler/stz-defs-db-serializer.stanza
```diff
@@ -23,31 +24,61 @@ public public defenum SrcDefinitionSource :
...
+public defstruct PackageDefinition : 
+  name:Symbol
+  file-name:String
+  imports:Tuple<PackageImport>
+
+public defstruct PackageImport : 
+  name:Symbol
+  prefixes:Tuple<PackageImportPrefix>
+
+public defstruct PackageImportPrefix :
```
CuppoJava, line 50: [P] Thanks this is necessary, but it seems to be unused currently.
What is supposed to create this structure? I also recommend changing
the names and types to more closely match their original names in
`il-ir.stanza`. See the `IPrefix` struct in `il-ir.stanza`.

So:
```
  sources => names
  Tuple<Symbol> => False|Tuple<Symbol>
  prefix:Symbol => prefix:String
```

The `False` case for `names` is used when the user types:

  prefix => pre-

Which indicates to use a single prefix for absolutely everything. See
the `IPrefix` type in `il-ir.stanza`.

After you change `Tuple<Symbol>` to `False|Tuple<Symbol>`, you'll have
to copy over `read-opt` and `write-opt` from `stz-pkg.stanza` for the
serialization. See the usages of `opt<FileInfo>(...)` for examples on
how they are used.

public defstruct IPrefix :
  names: False|Tuple<Symbol>
  prefix: String
with:
  printer => true

compiler/stz-defs-db.stanza
```diff
@@ -19,27 +19,38 @@ defpackage stz/defs-db :
 ;============================================================
 ;=================== Driver Arguments =======================
 ;============================================================
-
 public defstruct DefsDbInput :
   proj-files: Tuple<String>
   platform: Symbol|False
```
CuppoJava: Don't allow this platform to be false. Fill it with OUTPUT-PLATFORM when this is created.

compiler/stz-defs-db.stanza
```diff
@@ -19,27 +19,38 @@ defpackage stz/defs-db :
...
+
+; Helper function to compile the input data into IL-IR.
+protected defn compile-input (input:DefsDbInput) -> [NameMap, Seqable<IPackage|Pkg>] :
   ;Compute platform
   read-config-file()
-  val db-platform = match(platform(input)) :
-    (s:Symbol) : s
-    (f:False) : OUTPUT-PLATFORM
-    
+  val db-platform =
```
CuppoJava: Move this logic out of here. See the comment beside `platform:Symbol|False`.

compiler/stz-defs-db.stanza
```diff
@@ -53,111 +64,128 @@ public defn defs-db (input:DefsDbInput, filename:String) :
...
+; Expected Behavior : 
+; - When encountering an `IVisibility` node, traverse its child using that node's visibility
+; - When encountering an `IBegin`, traverse its children with the current visibility.
+; - The initial visibility is `Private`
+; - If an expression is not `Indexable`, it is skipped
+; - If an `Indexable` does not have an `info`, it is skipped
+; - If an `Indexable`'s name cannot be found, it is skipped 
+;
+protected defn index-expressions (exps:Seqable<IExp>, nm:NameMap) -> Seq<Indexed> :
+  var print? = false
```
CuppoJava: Debug code.

compiler/stz-defs-db.stanza
```diff
@@ -53,111 +64,128 @@ public defn defs-db (input:DefsDbInput, filename:String) :
...
+  generate : 
+    defn loop (e:IExp, current-visibility:Visibility) :
+      match(e) : 
+        ; Case 1: We encounter a visibility node. Walk its
```
CuppoJava, line 119: I like the structure of the comments here. It makes both the intention and implementation of the algorithm very clear.

compiler/stz-defs-db.stanza
```diff
@@ -53,111 +64,128 @@ public defn defs-db (input:DefsDbInput, filename:String) :
...
+; - When encountering an `IVisibility` node, traverse its child using that node's visibility
+; - When encountering an `IBegin`, traverse its children with the current visibility.
+; - The initial visibility is `Private`
+; - If an expression is not `Indexable`, it is skipped
+; - If an `Indexable` does not have an `info`, it is skipped
+; - If an `Indexable`'s name cannot be found, it is skipped 
+;
+protected defn index-expressions (exps:Seqable<IExp>, nm:NameMap) -> Seq<Indexed> :
+  var print? = false
+  generate :
```
CuppoJava: `generate<Indexed>` for clarity.

CuppoJava (changes requested):
Fantastic work. Implementation is both more powerful and cleaner than before.

compiler/stz-defs-db.stanza
m-hilgendorf: This comes directly from the CLI command, so I added it to `stz-main.stanza`. It turns out that the previous implementation always gave `false` for `platform` and it's not immediately useful to support cross-compilation for indexing commands, so I hardcoded it to OUTPUT-PLATFORM.

compiler/stz-defs-db.stanza
m-hilgendorf: Removed! Thank you

compiler/stz-defs-db.stanza
m-hilgendorf: Fixed!

compiler/stz-defs-db-serializer.stanza
m-hilgendorf: I split out the construction of `PackageDefinition` objects into a helper called `index-definitions` that now appropriately handles the extraction of imports and prefixes. 

Thank you for the pointer on `read-opt` and `write-opt`!

compiler/stz-defs-db.stanza
m-hilgendorf: Removed!

PR #151, opened 2022-04-05
Fix naming for stz/test-bitset-intrinsics

jwatson0:
Fix for
```
stanza run-test build-stanza.proj tests/stanza.proj stz/stanza-tests
```
```
tests/stanza-tests.stanza:10.9: Could not find package stz/test-bitset-intrinsics, which is imported
by package stz/stanza-tests.
```


CuppoJava (approved):
Thanks!

PR #156, opened 2022-04-06
Fix a bug in skip-live

OlegPliss:
When preliminary check finds that partial GC cannot recover enough space for desired nursery size + allocation size., or  full GC is requested), `mark-compact` can be called for a heap with uninitialized area in the range from `old-objects-end` to `nursery-start`. In a very rare case when all old objects are live at this moment, and old objects end is moved to heap start, a scan for solid prefix can reach this area and try to decode garbage there. The fix is to eliminate a special mode of operation with old objects end set to heap start.

I can only reproduce this crash when setting desired max heap size below current heap size with `set-max-heap-size`.


CuppoJava (approved):
Thank you!

PR #157, opened 2022-04-07
Fix for 32-bit integer truncation in the interpreter

OlegPliss:
Some C compilers treat `int` as `int32_t`. In bit operations relative to `bitset_base` the index must be pointer-wide.


CuppoJava (approved):
Thanks!

PR #158, opened 2022-04-07
Implement set-max-heap-size

OlegPliss:
Implementation of the forgotten function `set-max-heap-size`. The function can shrink and grow the heap size up to `MAXIMUM-HEAP-SIZE`. Logical space of `MAXIMUM-HEAP-SIZE` is allocated for the heap when it is created and cannot be later extended. Memory is only committed to this space when necessary and cannot exceed the limit set by `set-max-heap-size`.

Current value of `MAXIMUM-HEAP-SIZE` is 8 GB.

### DIFF
--- a/compiler/cvm.c
+++ b/compiler/cvm.c
@@ -405,6 +405,7 @@ typedef struct{
   uint64_t* bitset;
   uint64_t* bitset_base;
   uint64_t size;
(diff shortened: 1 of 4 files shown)


CuppoJava (approved):
Thank you!

PR #160, opened 2022-04-22
Update core/Process library for Windows.

m-hilgendorf:
- Change `clib/retrieve_process_state` to accept a pointer to a `Process` instead of using `pid`.
- Remove usage of `OpenProcess`, as lookup by pid fails if called long after the process terminates.
- Return an error if `retrieve_process_state` fails on
Windows (either in `WaitForSingleObject` or `GetExitCodeProcess`)
- Add a `handle` field to `Process` to hold the process handle created by `CreateProcess`.

### DIFF
--- a/runtime/process-win32.c
+++ b/runtime/process-win32.c
@@ -44,26 +44,50 @@ static FILE* file_from_handle(HANDLE handle, FileType type) {
   return file;
 }
 
-int retrieve_process_state (Process* process, ProcessState* s, stz_int wait_for_termination) {
-  ProcessState state;
-  DWORD exit_code;
-
+// Polls a running `process` returning the `ProcessState`, optionally blocking until it has terminated.
+int retrieve_process_state (const Process* process, ProcessState* s, stz_int wait_for_termination) {
+  ProcessState state;  // the returned state. 
+  DWORD exit_code;     // the exit code of the process
+  HANDLE handle;       // the handle to the process 
+  
+  // By default, assume the process is still running.
   state = (ProcessState){PROCESS_RUNNING, 0};
 
-  if (wait_for_termination == 1) {
-    if (WaitForSingleObject(process->handle, INFINITE) == WAIT_FAILED) {
+  // First, attempt to open a handle to the process to query its state by process ID.
+  // This is required to check a currently running process, as we do not have the permissions
+  // associated with the original handle created by launch_process (??? TODO: verify the cause)
+  handle = OpenProcess(SYNCHRONIZE | PROCESS_QUERY_INFORMATION, FALSE, process->pid);
+
+  // If the wait_for_termination argument has been passed, and the handle is valid, we 
+  // call WaitForSingleObject until the process exits. Requires a valid handle.
+  if (wait_for_termination && handle) {
+    // if WaitForSingleObject fails, return an error. The caller must retrieve the
+    // platform error message and raise an exception.
+    if (WaitForSingleObject(handle, INFINITE) == WAIT_FAILED) {
       return -1;
     }
   }
 
-  if (!GetExitCodeProcess(process->handle, &exit_code)) {
+  // If the returned handle is invalid, it means that the process has failed and we need
+  // to query the return code using original handle to the process.
+  if (!handle) {
+    handle = process->handle;
+  }
+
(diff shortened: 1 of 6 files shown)


runtime/process-win32.c
```diff
@@ -242,7 +260,7 @@ stz_int launch_process(stz_byte* command_line,
       /* lpProcessAttributes  */ NULL,
       /* lpThreadAttributes   */ NULL,
       /* bInheritHandles      */ TRUE,
-      /* dwCreationFlags      */ 0,
+      /* dwCreationFlags      */ CREATE_NO_WINDOW,
```
m-hilgendorf, line 263: Small modification: this prevents a call to `Process` from opening a terminal window.

runtime/process-win32.c
```diff
@@ -44,34 +44,52 @@ static FILE* file_from_handle(HANDLE handle, FileType type) {
...
+  // By default, assume the process is still running.
   state = (ProcessState){PROCESS_RUNNING, 0};
 
-  process = OpenProcess(PROCESS_QUERY_INFORMATION | SYNCHRONIZE, FALSE, (DWORD)pid);
-  if (process == NULL) {
-    goto END;
+  // First, attempt to open a handle to the process to query its state by process ID.
+  // This is required to check a currently running process, as we do not have the permissions
+  // associated with the original handle created by launch_process (??? TODO: verify the cause)
+  handle = OpenProcess(SYNCHRONIZE | PROCESS_QUERY_INFORMATION, FALSE, process->pid);
```
CuppoJava, line 59: Can you double-check whether we need to worry about calling `CloseHandle` on the original `process->handle` at some point?

Currently, there is a little bit of asymmetry that I can't completely follow:
1) If line 59 succeeds, and `WaitForSingleObject` fails then neither the newly created handle nor the original `process->handle` is closed.
2) If line 59 succeeds, then the newly opened handle is closed at line 87, but `process->handle` is not closed.
3) If line 59 fails, then `process->handle` is closed. 
4) For all of the above if `GetExitCodeProcess` fails then neither handle is closed.

CuppoJava (commented):
Thank you for the very clean job. 

I am not familiar with the operation of `CloseHandle` and left one question. Can you let me know whether it's all okay, and I'll merge it?

runtime/process-win32.c
m-hilgendorf: Good catches, I understand the logic is a little wonky. 

I checked the docs on [CloseHandle](https://docs.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle)

- We should close `process->handle` if `process->handle != handle`. 
- If `GetExitCodeProcess` fails, we don't necessarily want to call `CloseHandle`, since it may have failed for a reason besides an invalid handle, in which case we don't want to invalidate the state of the `Process` that we called this function on. 
- We should call `CloseHandle(handle)` if `WaitForSingleObject` fails, but only if `GetLastError() != ERROR_INVALID_HANDLE`. While it might seem similar to the previous case, since the `handle` is only used locally in this function we should clean it up.

One of the tricky things is that `CloseHandle` may fail, which means if we call it in one of the error cases, we don't want the call to `GetLastError` to refer to the error that arose from `CloseHandle` when the exception gets thrown by the caller, as the error message will be incorrect.

m-hilgendorf: Something I'm not clear on is if we need to call `CloseHandle` at all. The docs are nebulous as to the impacts, it might just be a small memory leak.

CuppoJava: Okay. Thank you for reading into this. 

It seems, more or less, that `CloseHandle` is akin to `new/delete`. You need to call it once and exactly once for every handle that you create. 

In addition to a memory leak, you'll also have a resource leak if you don't do it. 

Here's something from the doc:

> If the function succeeds, be sure to call the [CloseHandle](https://docs.microsoft.com/en-us/windows/desktop/api/handleapi/nf-handleapi-closehandle) function to close the hProcess and hThread handles when you are finished with them. Otherwise, when the child process exits, the system cannot clean up the process structures for the child process because the parent process still has open handles to the child process.

Can you polish up the `CloseHandle` logic a bit to ensure that we always close our handles appropriately?

PR #163, opened 2022-06-21
Fix a bug in relocate-liveness-trackers

OlegPliss:
Relocation of liveness trackers was written in assumption that GC preserves allocation order. So the trackers in the list are visited in the decreasing address order and the tracked value are always below the tracker. That was true until copying evacuation from the nursery was introduced. The evacuation can change the allocation order and so can break the assumptions. It may lead to subtle bugs with references in the tracker list and references from trackers to tracked values  are not updated.


CuppoJava (approved):
Thank you thank you thank you!

PR #164, opened 2022-06-22
Fix non-matching printf format specifiers in stz-asmjit.cpp

OlegPliss:
Use standard macros for format specifiers as described here: [https://en.cppreference.com/w/c/types/integer]


CuppoJava (approved):
Thanks!

PR #166, opened 2022-07-27
Index doc strings in definitions database/pkg files

m-hilgendorf:
- Added logic to extract IDoc from IL-IR in generation of definitions database from source code
- Added logic to TL and DL IRs to preserve doc strings when compiled into fpkg/pkg files.

### DIFF
--- a/compiler/stz-defs-db.stanza
+++ b/compiler/stz-defs-db.stanza
@@ -132,13 +132,14 @@ defn build-package-defs (pkg:Pkg, stamp:PackageStamp, nm:NameMap) -> PackageDefi
   ;Convert each of the exports into the database Definition objects.
   val definitions = to-tuple $     
     for e in exports-with-infos seq :
-      Definition(name, info, kind, visibility, annotation?) where :
+      Definition(name, info, kind, visibility, annotation?, documentation?) where :
         val name = name(id(rec(e)))
         val info = info(e) as FileInfo
         val visibility = visibility(e)
         val kind = kindof(rec(e))
         val annotation? = annotate?(e, nm)
-
+        val documentation? = documentation?(e)
+  
   ;Bundle the results in the PackageDefinitions datastructure.
   PackageDefinitions(name(pkg),
                      stamp,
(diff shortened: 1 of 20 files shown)


m-hilgendorf:
Since this changes the .pkg layout on disk the compiler has to be bootstrapped twice: first to compile the compiler with new pkg layout, and second to compile pkg files that use this layout.

compiler/stz-input.stanza
```diff
@@ -303,20 +303,27 @@ defn split-packages (e:IExp, default-imports:Tuple<IImport>) :
...
+  let loop (doc?:IDoc|False = false) : 
+    if not empty?(eseq) :
+      val pkg? =
+        match(peek(eseq)) :
+          (d:IDoc) :
+            next(eseq) 
+            loop(d)
+          (e:IDefPackage) :
+            next(eseq)
+            val exps = to-list(take-while({_ is-not IDefPackage}, eseq))
```
CuppoJava, line 315: I will fix this during the merge. Leaving a note for future purposes.

This algorithm won't work if there are multiple packages defined in the same file. 

E.g. 
```
IDoc
IDefPackage
... other expressions ...
IDoc
IDefPackage
... other expressions ...
```

The second IDoc will get swallowed up.

compiler/stz-tl-ir.stanza
```diff
@@ -211,7 +211,8 @@ public defast :
          lbl: Symbol
       TLInit :
          comm: LSComm
-
+      TDoc :
+         string: TExp
```
CuppoJava, line 215: Changed to a simple `String` because the extra wrapped `TExp` is not useful to us.

compiler/stz-tl-to-dl.stanza
```diff
@@ -100,17 +100,26 @@ defn compute-exports (p:TPackage, namemap:NameMap) :
...
       else :
         [Private, false]
-    add(exports, Export(n, vis, r, info))
-    
+    add(exports, Export(n, vis, r, info, d?))
+  
+  ; Track the most recently observed doc string and associate
+  ; it with the next export. Intermediate doc strings are dropped.
+  var last-doc? : False|String = false
+  defn export (n:Int, r:Rec) :
```
CuppoJava, line 115: Changed for last-doc to be tracked explicitly during the loop instead of implicitly via the mutated variable.

This is to accomplish two behavioural changes:
- TDefTuple carries the same doc string for all its definitions.
- Doc strings that are not attached to definitions are thrown away, and don't get spuriously assigned to the next definition.

compiler/stz-defs-db.stanza
```diff
@@ -193,9 +195,14 @@ protected-when(TESTING) defn build-package-defs (ipackage:IPackage, stamp:Packag
 ; - If an `Indexable` does not have an `info`, it is skipped
 ; - If an `Indexable`'s name cannot be found, it is skipped 
 protected-when(TESTING) defn index-expressions (exps:Seqable<IExp>, nm:NameMap) -> Seq<Indexed> :
-  generate<Indexed> : 
+  generate<Indexed> :
+    var documentation?:False|String = false
```
CuppoJava, line 199: Changed this to carry documentation? within the loop arguments itself. 

Changed behaviour to match algorithm in indexing TL-IR. IDoc expressions preceding expressions that do not correspond to a definition are thrown away.

PR #167, opened 2022-08-03
Added -check flag to compile commands

m-hilgendorf:


compiler/stz-compiler-main.stanza
```diff
@@ -18,6 +18,10 @@ compilation flows:
   3) We are compiling in unoptimized mode and only need to compile to
   .pkg files.
 
+  4) We are compiling in unoptimized mode and are running "check"
+
+  5) We are compiling in optimized mode and are running "check"
+
```
m-hilgendorf, line 24: These two cases seem to be equivalent.

compiler/stz-compiler-main.stanza
```diff
@@ -166,19 +171,26 @@ public defn compile (proj-manager:ProjManager,
     val output-pkgs = Vector<FileStamp>()
     
     ;Save optimized packages if necessary
-    if pkg-dir is String and optimize? :
+    if pkg-dir is String and optimize? and (not check?) :
```
m-hilgendorf, line 174: If we're checking, we don't want to write any additional output.

PR #179, opened 2022-11-28
Update tests/conantest02/Makefile to work with released 0.17.42

jwatson0:
Small tweak to Makefile.  Just running `make` in the `tests/conantest02` will build the conan asmjit sample test in isolation with released version `0.17.42`.


CuppoJava (approved):
Thanks!

